PR #19, opened 2015-01-08
Allow "extract" outside original image bounds.

### LINKED ISSUE #18: ValueError: shape mismatch in `ocropus-gpageseg`
When I run `ocropus-gpageseg` on the following image:

0001 bin

I get a `ValueError` which prevents any lines from being extracted:

```
$ ocropus-gpageseg -n --minscale 10 --maxcolseps 0 book-703662b.crop/0001.bin.png

########## /usr/local/bin/ocropus-gpageseg -n --minscale 10 --maxcolsep

book-703662b.crop/0001.bin.png
scale 13.8564064606
computing segmentation
computing column separators
computing lines
propagating labels
spreading labels
number of lines 4
finding reading order
writing lines
1: (slice(3L, 32L, None), slice(3L, 607L, None))
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-gpageseg", line 435, in safe_process1
    process1(job)
  File "/usr/local/bin/ocropus-gpageseg", line 408, in process1
    binline = psegutils.extract_masked(1-cleaned,l,pad=args.pad,expand=args.expand)
  File "/usr/local/lib/python2.7/site-packages/ocrolib/toplevel.py", line 213, in argument_checks
    result = f(*args,**kw)
  File "/usr/local/lib/python2.7/site-packages/ocrolib/psegutils.py", line 114, in extract_masked
    line = where(mask,line,amax(line))
ValueError: shape mismatch: objects cannot be broadcast to a single shape
```

danvk:
The problem was that `y0-pad` and `x0-pad` in `psegutils.extract_masked` could become negative. Chaos ensued.

This pads the output of `extract` with white pixels when you try to extract a region that goes outside the original image.

Fixes #18


tmbdev:
Hi, thanks. That looks like a reasonable fix, but I'm concerned that extract now can return images that are of a different size than what the caller expects. I'm not sure whether that is going to break code somewhere. It's probably better to just return the exact size subimage that was requested but pad it.

Doing that directly is sort of fiddly, but it's actually easy to do with scipy.interpolation.affine_transform; you can give that negative offsets, desired output width and height (output_shape), and boundary conditions; just use an identity matrix for the matrix. order should be 0 or 1. Do you think you could change your patch to use affine_transform instead?

(I'd do it myself, but I'm busy bug fixing the clstm implementation right now.)

danvk:
`extract` will always return images that are `(y1-y0)x(x1-x0)`, which wasn't the case before.

I can take a crack at using `scipy.interpolation.affine_transform`.

danvk:
I've reworked the code to use `affine_transform` and added a test image which reproduces the problem. Let me know if this looks right.

danvk:
ping

tmbdev:
Thanks; sorry, I was traveling. Thanks for looking into the patch. You still special-case interpolation.shift in that patch. The nice thing about affine_transform is that it doesn't need any special cases for any shifts. If you know what I mean, maybe you could simplify the patch to just use affine_transform for everything (mostly deleting a few lines of code), or I can do it soon.

PR #22, opened 2015-01-09
Update ocropus-rtrain

overstable:


tmbdev:
(The latest C++ version is not a drop-in replacement anymore, however, so this won't currently work.)

PR #28, opened 2015-02-21
Patch for OS X / clang

Inndy:
Compile flag `-fopenmp` is for `gcc`, but not works on `clang`


tmbdev:
Thanks. Down the road, it would be good to move away from native.py

PR #37, opened 2015-05-13
Travis CI builds

QuLogic:
A very simple setup for testing via Travis. It uses Miniconda to install dependencies and then runs the test script. I originally tried a straight install, but compiling SciPy took so long, it [ran out of build time](https://travis-ci.org/QuLogic/ocropy/builds/62312903). The conda install is much much quicker.

You will of course have to enable Travis for your repo before this has a real effect. And it doesn't really check whether the result correct, only that something is produced, but that's how the existing test script is. One could also do coverage testing using the other script and coveralls.io, but that's something for a another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,10 +40,19 @@ Alternatively, dependencies can be installed into a [Python Virtual Environment]
     $ wget -nd http://www.tmbdev.net/en-default.pyrnn.gz
     $ mv en-default.pyrnn.gz models/
 
+An additional method using [Conda](http://conda.pydata.org/) is also possible:
+
+    $ conda create -n ocropus_env python=2.7
+    $ source activate ocropus_env
+    $ conda install --file requirements.txt
+    $ wget -nd http://www.tmbdev.net/en-default.pyrnn.gz
+    $ mv en-default.pyrnn.gz models/
+    $ python setup.py install
+
 To test the recognizer, run:
 
     $ ./run-test
-    
+
 Running
 -------
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ trace by default since it seems to confuse too many users).
 Installing
 ----------
 
+[![Build Status](https://travis-ci.org/tmbdev/ocropy.svg)](https://travis-ci.org/tmbdev/ocropy)
 [![Join the chat at https://gitter.im/tmbdev/ocropy](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/tmbdev/ocropy?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
 To install OCRopus dependencies system-wide:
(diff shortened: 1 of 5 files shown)


tmbdev:
Thanks; this is very useful. I'll try to merge it.

Sorry this has taken so long; I've been focusing on CLSTM.

QuLogic:
I rebased, but master is [currently broken](https://travis-ci.org/QuLogic/ocropy/builds/84731314).

pdessauw:
I have the same error which I corrected by downgrading the Pillow library:
`pip install -U "Pillow<3"`

Another library problem is happening with newer version numpy (meaning that you have to downgrade scipy in the same time): I use **numpy 1.9.2** and **scipy 0.15.1**.

Maybe it can be included in the new requirements files (?).

tmbdev:
Travis-CI is very useful, but there are other changes in this patch that change the search path etc. Could you perhaps separate those out and also explain? For example, why do you want to remove searching in /usr/... directories for models?

QuLogic:
It doesn't make sense to search in `/usr/local` if the package is not installed there; on Travis it's installed in the conda environment. The change makes it search relative to the installed location instead.

kba:
@QuLogic: I removed the model-search commit (merged in #152), rebased against current master and adapted the travis.yml to current `requirements.txt` in https://github.com/kba/ocropy/tree/travis-conda. Travis check is running smoothly https://travis-ci.org/kba/ocropy/builds/176224957 :+1:

If that looks alright to you, please force-push those changes to your `travis-conda` branch, so we can at long last merge this :-)

kba:
I just realized the README.md needs an update too now with the changed requirements.

QuLogic:
I just rebased this myself, coincidentally...

README.md
```diff
@@ -40,10 +40,20 @@ Alternatively, dependencies can be installed into a [Python Virtual Environment]
     $ wget -nd http://www.tmbdev.net/en-default.pyrnn.gz
     $ mv en-default.pyrnn.gz models/
 
+An additional method using [Conda](http://conda.pydata.org/) is also possible:
+
+    $ conda create -n ocropus_env python=2.7
+    $ source activate ocropus_env
+    # PyPI and conda disagree on the name of the PyTables package:
```
kba: The comment is obsolete now I think, though it is still true of course ;)

And can you use a fenced code block and skip the `$`? Then users can copy and paste the whole thing.

```sh
conda create -n ocropus_env python-2.7
source activate ocropus_env
...
```

QuLogic: Oops yes. All the other ones use `$` though.

kba: Good point about the rest of the README. It's a matter of taste I guess, not important.

zuphilip:
Thank you very much @QuLogic !

We will try to redirect the status of the badge, since Travis still needs to be approved here by @tmbdev .

zuphilip:
FYI Travis is activated and approved now.

PR #42, opened 2015-06-04
Add en-default model

danvk:


tmbdev:
Unfortunately, the models are a bit large (that's just because they contain log information). I'll try to make smaller models and put them into the repository in the future.

zuphilip:
Is http://www.tmbdev.net/ocropy/en-default.pyrnn.gz (13-Jan-2015 13:35) really newer than http://www.tmbdev.net/en-default.pyrnn.gz (02-Nov-2014 11:54)?

tmbdev:
I think they are the same model. I kept both because I didn't want to break old download scripts.

zuphilip:
Yes, they are the same, I checked their MD5 checksum.

aayushee:
I have tried downloading the default model from both the above links but I am still getting an error that it is not in gzip format. Can someone please suggest a solution for this?

zuphilip:
You don't need to do anything with the file, just put it into the `models` directory of your copy of `ocropy`. Does that work?

> I am still getting an error that it is not in gzip format

What do you do that causes this error? Which program is giving you an error?

aayushee:
Thanks for the reply!
It is working now. Somehow something was wrong with the model I downloaded earlier or my set up. 
I have another question though. Do you know the difference between ocropus-rtrain and ocropus-ltrain files? I have been using the rtrain file but was wondering what is the difference between the two. Is it that one uses RNN and the other uses LSTM for training?

amitdo:
@aayushee, I suggest to open a new issue for your new question.

PR #53, opened 2015-07-27
Added better print methods for distinguishing between info and error …

lunactic:
I changed printing to a new (future proof) method and changed the output a bit. 
My reasoning behind this is the following: 

I am currently writing a nodejs wrapper to build a RESTful API to use ocropy over the web. To be able to easily distinguish if an operation ran correct or not I need to check if ocropy wrote to stderr or not.

If you can agree with these changes, I could add the same printing to the other ocropy methods as well.

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -194,9 +194,9 @@ def pil2array(im,alpha=0):
 def array2pil(a):
     if a.dtype==dtype("B"):
         if a.ndim==2:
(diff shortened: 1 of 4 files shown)


tmbdev:
Sorry it took so long; I've been focusing on CLSTM.

lunactic:
No worries :)

Should I make the same changes in the other python files as well?

PR #68, opened 2015-11-19
smaller vertical separators

jze:
I work a lot with bordered tables and observed damaged characters near the left and right cell borders. I think the vertical separators are expanded too much.


tmbdev:
That looks useful, but could you please make these values into command line parameters? I'm reluctant to change them because it would require extensive testing to figure out how that affects other page segmentation (since I currently don't have a setup for testing this easily on large datasets).

Something like:

    d1 = int(max(5,args.separator_multiplier_h * scale)) + args.sepwiden
    seps = minimum(thresh,maximum_filter(grad,(int(args.septhresh_multiplier_h*scale),int(5*scale))))

The defaults should just be 1.0

Eventually, this code will get replaced by trainable 2D LSTM models.

kba:
Superseded by #138

PR #65, opened 2015-11-08
Fix casting rule issue

dblN:
With `numpy 1.10.1`, I got an error below.

```
INFO:  
INFO:  ########## ./ocropus-gpageseg book/0001.bin.png
INFO:  
INFO:  book/0001.bin.png
INFO:  scale 41.8090899207
INFO:  computing segmentation
INFO:  computing column separators
INFO:  computing lines
Traceback (most recent call last):
  File "./ocropus-gpageseg", line 423, in safe_process1
    process1(job)
  File "./ocropus-gpageseg", line 379, in process1
    segmentation = compute_segmentation(binary,scale)
  File "./ocropus-gpageseg", line 314, in compute_segmentation
    seeds = compute_line_seeds(binary,bottom,top,colseps,scale)
  File "./ocropus-gpageseg", line 263, in compute_line_seeds
    bmarked *= (bottom>t*amax(bottom)*t)*(1-colseps)
TypeError: Cannot cast ufunc multiply output from dtype('int32') to dtype('bool') with casting rule 'same_kind'
```

It should avoid violating the `same_kind` rule.

PR #75, opened 2015-12-31
removed augmented assignment in ocropus-gpageseg

knoxdw:
... to avoid numpy type casting errors. 

Came across the same errors mentioned in Ocropy's Google Groups forum in October 2015:
https://groups.google.com/d/msg/ocropus/U4_MNVRMKxA/b7iwyKbEBQAJ

Apparently numpy doesn't handle casting in augmented assignment in the same way that it does when multiplying on the right hand side of the assignment.


amitdo:
Seems like a duplicate of #65.

Issue #84, opened 2016-03-01
[ocropus-gtedit] "Save page as" defaults to "Correction"


kba:
Since the title of the HTML page is hard-coded ot `Correction` and Firefox uses the page title as the default proposal for its "Save page as" functionality, it will always try to save to a file `Correction.html`.

zuphilip:
Fixed in a84fdf1

PR #85, opened 2016-03-01
[gtedit] Use HTML file name instead of "Correction", #84

kba:
Now, Firefox (and probably other browsers) will propose to use the
actual filename as the template for "Save page as".


zuphilip:
Thank you!

PR #103, opened 2016-06-28
Fix for 'TypeError: float argument required, not str'

jmorrey:
We were trying to cast "scale" to a string then print it as a float.

### DIFF
--- a/ocropus-gpageseg
+++ b/ocropus-gpageseg
@@ -370,7 +370,7 @@ def process1(job):
         print_error("%s: bad scale (%g); skipping\n"%(fname,str(scale)))
         return
     if scale<args.minscale:
-        print_error("%s: scale (%g) less than --minscale; skipping\n"%(fname,str(scale)))
+        print_error("%s: scale (%s) less than --minscale; skipping\n"%(fname,str(scale)))
         return
 
     # find columns and text lines


zuphilip:
Here is an example for reproducing the error which is fixed here:
```
ocropus-gpageseg --minscale 12 --scale 10 tests/testpage.png
INFO:
INFO:  ########## /usr/local/bin/ocropus-gpageseg --minscale 12 --scale 10 tes
INFO:
INFO:  tests/testpage.png
INFO:  scale 10.0
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-gpageseg", line 423, in safe_process1
    process1(job)
  File "/usr/local/bin/ocropus-gpageseg", line 373, in process1
    print_error("%s: scale (%g) less than --minscale; skipping\n"%(fname,str(scale)))
TypeError: float argument required, not str
```

kba:
Why cast `scale` to `str` in the first place? The point of using `"%g" % (123)` or similar is to do explicit to-string conversion.

```python
# Why this
print_error("%s: scale (%s) less than --minscale; skipping\n"%(fname,str(scale)))
# instead of this
print_error("%s: scale (%g) less than --minscale; skipping\n"%(fname,scale))
```

This happens in quite a few places:

```sh
ag -l 'print.*strc?\('
ocropus-nlbin
ocropus-gpageseg
ocropus-rpred
ocrolib/toplevel.py  
```

But that is a different issue and should be fixed in a new PR.

zuphilip:
Thank you very much, @jmorrey, for reporting and fixing the issue with this pull request!

Issue #104, opened 2016-07-17
hOCR document metadata: old version hard-coded


kba:
`ocr-system` is is [hard-coded to `ocropus-0.4`](https://github.com/tmbdev/ocropy/blob/master/ocrolib/hocr.py#L18) (should be `-1.0`).

Also, it's `ocropus-0.4` and `OCRopus` not `ocropy` or `OCRopy` (#97)

PR #105, opened 2016-07-17
hocr metadata: bump version 0.4 -> 1.0, ocropus -> ocropy

kba:
See #104

### DIFF
--- a/ocrolib/hocr.py
+++ b/ocrolib/hocr.py
@@ -14,8 +14,8 @@ header_template = """\
 <head>
 <title>OCR Results</title>
 <meta http-equiv="content-type" content="text/html; charset=utf-8" />
-<meta name="Description" content="OCRopus Output" />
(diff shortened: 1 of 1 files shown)


ocrolib/hocr.py
```diff
@@ -14,8 +14,8 @@
 <head>
 <title>OCR Results</title>
 <meta http-equiv="content-type" content="text/html; charset=utf-8" />
-<meta name="Description" content="OCRopus Output" />
-<meta name="ocr-system" content="ocropus-0.4" />
+<meta name="Description" content="ocropy Output" />
```
zuphilip: Do we need the `Description` here? Is this the same information as the `title`? Should this be `DC`-prefixed?

--> My suggestion is to simply delete this `meta`-tag with `Description`. Also Tesseract does not have such a line.

zuphilip:
Thank you!

PR #108, opened 2016-08-30
firefox is not needed for ocropy to be installed

kba:
Installing firefox pulls in a lot of dependencies that are not needed in a headless setup.


zuphilip:
Thank you!

PR #109, opened 2016-08-30
opencv is used in the OLD scripts but not for the installed tools

kba:
This pulls in a lot of graphics dependencies that won't be used in the installed tools.


zuphilip:
Thank you, @kba!

PR #110, opened 2016-08-30
python setup.py install should not fail without models

kba:
While it's sensible to remind the user to install a model and have available models at build time installed to /usr/share/ocropus, it should be possible to install ocropy without any models. They can always be provided via CLI flag.

This patch will install **all** `models/*.pyrnn.gz` files that are found, not just the default. But if none are found, it will still succeed.


zuphilip:
Thank you!

PR #114, opened 2016-10-07
[DISCUSSION] Py3k simple

zuphilip:
This contains the commits from https://github.com/QuLogic/ocropy/tree/py3k-simple by @QuLogic. I separate them from the commits for Travis integration and rebase them to current master. Thus, we can discuss this here.

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -493,7 +493,7 @@ def unpickle_find_global(mname,cname):
     if mname=="lstm.lstm":
         from . import lstm
         return getattr(lstm,cname)
-    if not mname in sys.modules.keys():
+    if not mname in sys.modules:
         exec("import "+mname)
     return getattr(sys.modules[mname],cname)
 
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -3,7 +3,7 @@
 ### common functions for data structures, file name manipulation, etc.
 ################################################################
 
-from __future__ import print_function
+from __future__ import absolute_import, print_function
 
 import os,os.path
 import re
@@ -15,14 +15,13 @@ import inspect
 import glob
 from numpy import *
 from scipy.ndimage import morphology
-import ligatures
 import multiprocessing
-import lstm
 import pylab
-
 from pylab import imshow
-import morph
-from toplevel import *
+
+from . import ligatures
+from . import morph
+from .toplevel import *
 
 ################################################################
 ### exceptions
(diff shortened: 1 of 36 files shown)


ocrolib/extras/cairoextras.py
```diff
@@ -34,14 +36,14 @@ def create_cairo_font_face_for_file(filename, faceindex=0, loadoptions=0):
     cairo_t = PycairoContext.from_address(id(cairo_ctx)).ctx
     _cairo_so.cairo_ft_font_face_create_for_ft_face.restype = ctypes.c_void_p
     if FT_Err_Ok != _freetype_so.FT_New_Face(_ft_lib, filename, faceindex, ctypes.byref(ft_face)):
-        raise "Error creating FreeType font face for " + filename
+        raise Exception("Error creating FreeType font face for " + filename)
```
QuLogic: I should probably make this a more specific error.

ocrolib/common.py
```diff
@@ -493,7 +493,7 @@ def unpickle_find_global(mname,cname):
     if mname=="lstm.lstm":
         from . import lstm
         return getattr(lstm,cname)
-    if not mname in sys.modules.keys():
+    if not mname in sys.modules:
```
QuLogic: Should have made this `not in`.

stweil:
@zuphilip, I suggest breaking this into smaller PRs. Some of them are trivial (replacing tabs by spaces) or easy to review (print function) and could be applied soon. That would reduce the complexity a little.

zuphilip:
I just created a new "experimental" branch `python3` and changed the base in this PR here, because this might need more work and maintenance is currently focused on Python 2.7. I guess some more step-by-step approach might be easier to review. I am open for any PR if someone want to work on that.

kba:
These are easy to review and can be merged right away:
8cac18895da4d9cab4945d9662591cd40a9acacb (new style exceptions)
877891582af55df229d055f7b9ae953bb4795922 (remove tabs)

QuLogic:
TBH, I felt that all the commit are small enough that they could be reviewed in the commit view of the PR itself (except for the print function one.)

mittagessen:
I just want to mention that there is no way you're going to get models pickled using python 2.7 to deserialize on py3k without writing your very own pickle parser. You can work around most issues but ultimately ``SeqRecognizer`` is an old-style class which don't exist anymore on 3.x.

ChillarAnand:
This is [meant to be separate branch rather](https://github.com/tmbdev/ocropy/pull/206#issuecomment-316143805) than providing both Py2 & Py3 support. May be we should add a note to readme.

Issue #113, opened 2016-10-07
Swapped text blocks order


PedroBarcha:
I've notice that OCRopus frequently swap the order of the texts recognized. In the folowing image, for instance, "LIBRARY" and "May 27,1993" were correctly recognized, but their position isn't correct in the output:
![8501_001](https://cloud.githubusercontent.com/assets/11379168/19176339/215fd584-8c15-11e6-8524-2e602fb524e6.png)

**OCROpus output:**
UNIVERSITY OF NEVADA, LAS VEGAS
45O5 MARYLAND PARKWAY LAS VEGAS, NEVADA 89154-7001 (702) 7393286
YUCHENG LIU
5613 S EASTERN AVE
LAS VEGAS NV 89119
DEAR LIU:
LiBRARY
May 27, 1993
We are pleased to inform you that you have been selected for a
carrel assignment for the fall semester. Please come to the
Circulation Desk to pick up your key and get your room assignment
at the beginning of the fall semester. If you are already
currently assigned to a carrel and have a key, you may be able to
stay where you are. If I do need to move you for any reason, I
will let you know. For instance, if your carrel happens to be one
of the ones scheduled to have the locks removed, I will have to
change your carrel assignment.
Sincerely,
--) . ----'e - AC
Sheila Beard
Circulation Section

PedroBarcha:
Why does it happen???

Another example ("September 20, 1993" is in the wrong position):
![8520_001](https://cloud.githubusercontent.com/assets/11379168/19176412/d585d82e-8c15-11e6-85d2-b34aad9ebcc2.png)

**OCRopus output:**
MY. Kevin Grover
Programmer
University of Nevada Las Vegas
i505 South Maryland Parkway
Las Vegas, NV 89154
Dear Mr. Grover:
w 0 b
September 20, 1993
Do you want to hear some EgS9 news?
'-, .,
: sD -
Twenty- three months ago we scheduled a very special person, Dr. Denis Waitley,
on our calendar. If you know of him, you will probably want to skip the rest
of this letter and simply call for your tickets to the November 16 University
of Nevada, Las Vegas ''Lessons in Leadership'' program here in Las Vegas.
If you don't yet know of Denis Waitley, look at what others have said:
''Denis Waitley has always been one step ahead of all of us. .
his new program, he leapfrogs even further. .
all of us. .
. With
. . Denis is a mentor for
. . This is special stuff1''--Pat Riley, Head Coach of the
New York Knicks
[...]

PR #118, opened 2016-10-13
Changed reading order algorithmus slightly

zuphilip:
This will solve #113 . I tried all examples in the test they look pretty good. But I guess we should do more tests and maybe @tmbdev can also look at this PR. I don't understand exactly the previous algorithms or the "separates" method.

The now proposed algorithm distinguishes more cases and can therefore make a (hopefully) better decision whether to look horizontally or vertically. Here are the different case which should help to understand the algorithm:

reading-order


zuphilip:
The current algorithm for the reading order is described in this paper http://www.dfki.de/web/research/publications/renameFileForDownload?filename=HighPerfDocLayoutAna.pdf&file_id=uploads_552

PR #123, opened 2016-10-20
nlbin: fix VisibleDeprecationWarning about non-integers

kba:
ocropus-nlbin causes

```
/usr/lib/python2.7/dist-packages/numpy/core/numeric.py:190: VisibleDeprecationWarning: using a non-integer number instead of an integer
 will result in an error in the future
  a = empty(shape, dtype, order)
```

and tries to suppress it:

```python
        v = morphology.binary_dilation(v,structure=ones((e*50,1)))
```

`e` is a float, so the first tuple element will also be a float. But it should be an integer.

numpy 1.11.0 at least outputs a `VisibleDeprecationWarning` which won't be filtered.

### DIFF
--- a/ocropus-nlbin
+++ b/ocropus-nlbin
@@ -164,7 +164,7 @@ def process1(job):
 
     # estimate low and high thresholds
     if args.parallel<2: print_info("estimating thresholds")
-    warnings.filterwarnings("ignore", category=DeprecationWarning)
+    warnings.filterwarnings("ignore", category=VisibleDeprecationWarning)
     d0,d1 = flat.shape
     o0,o1 = int(args.bignore*d0),int(args.bignore*d1)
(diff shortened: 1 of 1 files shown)


QuLogic:
A correct shape should be passed instead of ignoring the warnings.

kba:
I fixed the source of the warning in https://github.com/tmbdev/ocropy/pull/123/commits/4493db77c4f231f775b8411cc0e71b7adbf81027. Not sure whether `math.floor` better fits current behavior.

zuphilip:
@QuLogic : Does this looks fine for you now?

QuLogic:
Looks good to me, pretty similar to what I did for kraken, now that I think about it (I knew it looked familiar for some reason.)

zuphilip:
Thank you @kba for the fix and @QuLogic for the review!

Issue #127, opened 2016-10-24
Check that input is not color image fails


zuphilip:
It looks that the current check, whether an input picture is just black/white or using more colors is not working as expected. This seems to be the case in `ocropus-rpred`, see also #55, but I guess there may be other similar cases in other scripts.

Here is an colorful line for inputing in `ocropus-rpred`, which should yield to an error message about the colors:
![color-line](https://cloud.githubusercontent.com/assets/5199995/19662115/991469bc-9a36-11e6-979b-55fc32e88040.png)
which will not produce an error message:
```
$ ocropus-rpred tests/color-line.png
INFO:
INFO:  ########## ./ocropus-rpred tests/color-line.png
INFO:
INFO:  #inputs: 1
# loading object /usr/local/share/ocropus/en-default.pyrnn.gz
INFO:  tests/color-line.png:= 's<= s
```

kba:
The color check will always fail because [ocrolib.read_image_gray](https://github.com/tmbdev/ocropy/blob/master/ocrolib/common.py#L241-L242) already catches that:

```python
    if a.ndim==3:
        a = mean(a,2)
    return a
```

So it's always a 2-dimensional array of grayscale float values but still not necessarily a binary image. To check for that, a must only contain two unique values `0` and `1`.

Issue #55, opened 2015-09-03
ValueError: setting an array element with a sequence.


braddockcg:
On some images, rpred and rtrain's dewarp function throws an error because the padding is insufficient.  This seems to be particularly true of images which have borders or other noise near the top and bottom edges.  Attached is an example.

![ocropy-bug-example](https://cloud.githubusercontent.com/assets/1967740/9670848/877f2cb4-5243-11e5-9319-636579225aac.png)

I have a simple, but unsatisfying, fix for this where I double the amount of padding.  See my github fork commit (branch bugfix) at:
https://github.com/braddockcg/ocropy/commit/be2e6d1e996b75898ff0958fc680c995b2574e4d

The full error is:

Traceback (most recent call last):
  File "/usr/local/bin/ocropus-rpred", line 245, in safe_process1
    return process1(arg)
  File "/usr/local/bin/ocropus-rpred", line 145, in process1
    line = lnorm.normalize(line,cval=amax(line))
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lineest.py", line 59, in normalize
    dewarped = self.dewarp(img,cval=cval,dtype=dtype)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lineest.py", line 56, in dewarp
    dewarped = array(dewarped,dtype=dtype).T
ValueError: setting an array element with a sequence.

tmbdev:
Thanks; I'll look into this again soon, when I get around to cleaning up some of this code.

zuphilip:
The picture is a color image and it should be just a black/white picture. If you are binarize the picture, then the `ocropus-rpred` is working as it should. It looks that the check, which should prevent that case, is not working correctly.

zuphilip:
While my comment above is true, it seems that this is more like some coincident and a real fix should somehow increase the padding. I created now another PR #133 extending your idea, which will solves this issue then.

zuphilip:
Closed by #133

PR #129, opened 2016-10-25
Raise a ValueError for non-binarized images, #127 #55

kba:
Simple checks to ensure that images loaded for training / recognition are black and white.

### DIFF
--- a/ocropus-rpred
+++ b/ocropus-rpred
@@ -133,6 +133,8 @@ def process1(arg):
     (trial,fname) = arg
     base,_ = ocrolib.allsplitext(fname)
     line = ocrolib.read_image_gray(fname)
+    if not ocrolib.isbinarized(line):
+        raise ValueError("Image %s is not binarized" % (fname))
     raw_line = line.copy()
     if prod(line.shape)==0: return None
     if amax(line)==amin(line): return None
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -216,6 +216,11 @@ def isintarray(a):
 def isintegerarray(a):
     return a.dtype in [dtype('int32'),dtype('int64'),dtype('uint32'),dtype('uint64')]
 
+def isbinarized(a):
+    """Tests whether an array has only two unique values 0 and 1"""
+    a = sorted(unique(a))
+    return len(a) == 2 and a[0] == 0 and a[1] == 1
+
 @checks(str,pageno=int,_=GRAYSCALE)
 def read_image_gray(fname,pageno=0):
     """Read an image and returns it as a floating point array.
--- a/ocropus-lpred
+++ b/ocropus-lpred
@@ -80,6 +80,8 @@ for trial in range(len(inputs)):
         fname = inputs[trial]
         base,_ = ocrolib.allsplitext(fname)
         line = ocrolib.read_image_gray(fname)
+        if not ocrolib.isbinarized(line):
+            raise ValueError("Image %s is not binarized" % (fname))
         line = preprocess(line)
         if line is None: continue
         outputs = array(network.forward(line))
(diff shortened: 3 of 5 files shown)


ocropus-rpred
```diff
@@ -133,6 +133,8 @@ def process1(arg):
     (trial,fname) = arg
     base,_ = ocrolib.allsplitext(fname)
     line = ocrolib.read_image_gray(fname)
+    if not ocrolib.isbinarized(line):
+        raise ValueError("Image %s is not binarized" % (fname))
```
zuphilip: I think here we should either replace the current (not working) check above or delete it.

zuphilip:
We potentially lose here the possibility to train, recognize grayscale text lines. However, I think this didn't ever work properly as the example in #55 shows and maybe some code is just not thought about non-binarized text-lines. Do you agree? Then, this PR will not change backwards compatibility.

ocropus-rpred
kba: The check above checks whether the image is a rank 3 array of floats which won't work because read_image_gray already catches that case. Here we check that an array, regardless of rank, contains only 0 and 1. However you need both to check both rank and values because an RGB image that only contained e.g. pure red (1,0,0) and pure green (0,1,0) would also be considered "binarized".

kba:
> We potentially lose here the possibility to train, recognize grayscale text lines.

We could just omit the check in rtrain/ltrain or add/extend a --check/--nocheck flag. Do people train on non-binarized images?

zuphilip:
It was written in #55 that the problem also occurred in rtrain, but it seems that ltrain and lpred are not using dewarp functionality. I don't know whether people do train on non-binarized images with ocropy...

ocropus-rpred
zuphilip: But... `image.shape` is normally just the pair with the height and width. Does the `image.shape` for colorful pictures really contain more values?

zuphilip:
Okay, here is my suggestion: We will just do your checking in rtrain and rpred, because there we have an error report; and don't change lpred and ltrain. Okay?

kba:
I'm not even sure whether #55 is due to this, I'll investigate that further and ping back. Still I would find it useful to get a warning about non-binarized images regardless because in my experience, it was always an error.

uvius:
You may not be aware of it but training and recognition on gray line images (*.nrm.png instead of *.bin.png, which you get by running ocropus-gpageseg --gray) does give better  (sometimes dramatic: 20% error rate vs. 4% error rate) results, especially in cases where binarization leads to a loss of connectedness in glyph shapes. Forcing ocropy to only work on binary images is a step back in its usefulness. Please reconsider your changes.

ocropus-rpred
zuphilip: Okay, I have seen that this works as it should.

zuphilip:
Okay, we also discussed more. There might be other fixes for #55, I am experimenting a little more on that.

tmbdev:
From memory... The logic about binary vs grayscale is a bit complicated because some steps need binary and others need grayscale. Also, ocropus-rtrain was intended to be used with grayscale, but ocropus-rpred was not.

kba:
To be clear: I did not want to merge this as such. Feedback has been really useful, grayscale vs. binary should be documented as expected by different steps in the process and maybe add warnings in rpred.

amitdo:
>Also, ocropus-rtrain was intended to be used with grayscale, but ocropus-rpred was not.

So you can train it on grayscale but you can't use the model to do OCR on grayscale images? Then what's the point to train it on grayscale?

kba:
> especially in cases where binarization leads to a loss of connectedness in glyph shapes

Just wanted to highlight this from @uvius [comment](https://github.com/tmbdev/ocropy/pull/129#issuecomment-256770607), that seems like a good reason why to train on grayscale but recognize binary but I'm no expert.

zuphilip:
I created a new section in the FAQ about colorful, grascale and black-and-white pictures: https://github.com/tmbdev/ocropy/wiki/FAQ#when-can-colorful-grayscale-or-black-and-white-images-used , where I mostly copied some information together. @tmbdev I copied also your answer from the gitter to that page.

Can someone help to improve the answer in the [FAQ](https://github.com/tmbdev/ocropy/wiki/FAQ#when-can-colorful-grayscale-or-black-and-white-images-used)?

PR #139, opened 2016-11-04
Construct SVG output files by putting together the recognition resuts for each page

kba:
By @jze in #68

### DIFF
--- a/ocropus-svg
+++ b/ocropus-svg
@@ -1,0 +1,141 @@
+#!/usr/bin/env python
+
+import __builtin__ as python
+import random as pyrandom
+import sys,os,re,glob,argparse,codecs
+from pylab import median, imread
+
+import ocrolib
+
+parser = argparse.ArgumentParser("""
+Construct an SVG output file by putting together
+the recognition results for each page.
+You should usually invoke this program as 
+
+    ocropus-svg 'book/????.bin.png'
+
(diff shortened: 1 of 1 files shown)


zuphilip:
Here is an [example of the output](https://cloud.githubusercontent.com/assets/5199995/25781546/7f0b05a0-333b-11e7-9a9f-bb181a7ad4da.png) (converted into a png file for easier handling).

I think it is more interesting to continue hocrjs for visualizing the hocr output. Thus, I would suggest to not take this PR in the repo. @kba What do you think about this?

kba:
SVG is a nice format, portable, can be displayed in the browser and could also be scripted. I considered it for hocrjs but decided against it because hOCR is already HTML and easier to manipulate. If you want to share/compare a visual representation of the hOCR output, it seems like a good option. Github supports diffing SVGs, so you could use that to see what affects positioning etc.

zuphilip:
Okay, I see that there is some more value into this script. However, I am not sure we should have another top-level command `ocropus-svg` for that...

> If you want to share/compare a visual representation of the hOCR output, it seems like a good option.

Wouldn't we need then something like a hocr2svg tool? Which I guess is possible by rewriting the lines here and could be saved in the hocr-tools...

We could also think about integrating these svg lines into `ocropus-visualize-results`. These could actually be some replacement for the nowadays obsolete csegs parts in this script.

Issue #96, opened 2016-05-09
./ocropus-linegen: Import Error can be fixed with 'from PIL import ImageFont, ImageDraw'


anjueappen:
Got the following import error when running: 

> $ ./ocropus-linegen -h
> Traceback (most recent call last):
>   File "./ocropus-linegen", line 8, in <module>
>     import ImageFont,ImageDraw
> ImportError: No module named ImageFont

I am working in an anaconda virtual environment and running neither 'conda install pillow' nor 'pip install pillow --upgrade' appeared to help (although those were needed at first). So following pillow docs, I changed the following line in linegen
**-import ImageFont,ImageDraw
+from PIL import ImageFont,ImageDraw**

This appeared to work for me and I figured that I would double check if this is the correct fix before pushing any changes.

kba:
Related: #82

The second option, with the explicit `from PIL` is also what the [pillow docs](https://pillow.readthedocs.io/en/3.2.x/handbook/) recommend. In fact, I don't think that `import ImageFont,ImageDraw` will work with very old versions of pillow (< 1.0). Other users [have reported](https://github.com/tmbdev/ocropy/pull/37#issuecomment-153455244) needing to downgrade to pillow <= v.3.0

Maybe this is a good opportunity to discuss @QuLogic's [Travis CI](https://github.com/tmbdev/ocropy/pull/37) pull request again, since a continuous integration with a virtual environment like conda/virtualenv should help fix such issues.

zuphilip:
Should we take the commit https://github.com/tmbdev/ocropy/commit/122788917c39b851a3eb2e6026fa8e1ae3fde054 for fixing this?

I don't have an anaconda environment to test.

Agree on importance of CI.

kba:
> Should we take the commit 1227889 for fixing this?

:+1:

I'll look into CI again.

zuphilip:
This was fixed by #140 and work on CI started in #149.

PR #140, opened 2016-11-04
From PIL import

zuphilip:
This should solve #96


zuphilip:
I also deleted a duplicate import statement which was hidden in the main content below.

zuphilip:
Okay, I delete the last two commits form here and will create a new PR with them.

PR #142, opened 2016-11-05
Issue template, mostly for bug reports

kba:
An [issue template](https://help.github.com/articles/creating-an-issue-template-for-your-repository/) to make reporting bugs/problems/feature requests more structured.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -22,6 +22,8 @@
 
 ## Your Environment
 <!--- Include as many relevant details about the environment you experienced the bug in -->
-<!--- Run `git rev-parse HEAD` in the ocropy directory -->
-* Git revision:
+<!-- Run `python -V` -->
+* Python version:
+<!--- Run `git log -n1` in the ocropy directory -->
+* Git revision of ocropy: 
 * Operating System and version:
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,27 @@
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Expected Behavior
+<!--- If you're describing a bug, tell us what should happen -->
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+## Steps to Reproduce (for bugs)
+<!--- Provide the input image(s) and list the commands you entered. -->
+1.
+2.
+3.
+4.
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+<!--- Run `git rev-parse HEAD` in the ocropy directory -->
```
zuphilip: I have never used the `git rev-parse` command so far. Maybe, it is easier to suggest `git log -n 1` where the user might easier see if the version is outdated.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+* Git revision:
+* Operating System and version:
```
zuphilip: I think we should also include the python version here. The numpy, scipy, ... version could also be interesting, but I guess that is too much to ask for the default.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+<!--- Provide the input image(s) and list the commands you entered. -->
+1.
+2.
+3.
+4.
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+<!--- Run `git rev-parse HEAD` in the ocropy directory -->
+* Git revision:
```
zuphilip: Maybe add "of ocropy" here.

kba:
Okay, let's try this. If it turns out to be too complicated or not helpful we can always change/disable later.

zuphilip:
Thank you for working on this! Let us see now how this will be used.

PR #143, opened 2016-11-05
Clean up import statements

zuphilip:
These two import statements were hidden
in the main content below, but already
present in the beginning, i.e. they were
duplicate.

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -3,24 +3,31 @@
 ### common functions for data structures, file name manipulation, etc.
 ################################################################
 
-import os,os.path
+import os
+import os.path
 import re
-import numpy
-import unicodedata
 import sys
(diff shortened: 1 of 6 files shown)


zuphilip:
Split of from #140

zuphilip:
@kba Feel free to add to this branch more code cleaning commits.

zuphilip:
This looks ready now, right? I think, this should be fine to merge @kba.

kba:
I still have commits removing the `__all__ = ...`  in ocrolib/__init__.py and making the `import *` imports explicit but I'd prefer to test all the tools first. I'll post-pone those until CI and more tests.

zuphilip:
Okay, thank you for the commits here!

You can open a new PR whenever the other commits are ready and tested.

Issue #100, opened 2016-05-27
ocropus-rtrain : Fix & drastically improve --strip option (down to 3.4Mb)


e1000:
Hello there,

* First, there is a problem in the **ocropus-rtrain** option *--strip* : it strips the network only if the option is NOT passed (this is because of using the confusing 'store_false'  ..)

* Second, I've managed to trim down the network model to 3.4 Mb by deallocating many temporary variables.

But I can not make a pull request for this, because I had to fork & modify many things for my context ; however I isolated the commit (https://github.com/CBien/ocropy/commit/151a09dc269961b0c6362f67b0808493cd99144d), so that you can maybe cherry-pick or so

PR #144, opened 2016-11-05
strongly improved network stripping

zuphilip:
This is cherry-picking the commit by @CBien for discussion, see also issue #100.

### DIFF
--- a/ocropus-rtrain
+++ b/ocropus-rtrain
@@ -156,7 +156,8 @@ def save_lstm(fname,network):
         network.lstm.save(fname)
     else:
         if args.strip:
-            network.clear_log()
+            print yellow('saving stripped network (without temporary variables)...')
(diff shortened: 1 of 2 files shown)


ocropus-rtrain
```diff
@@ -156,7 +156,8 @@ def save_lstm(fname,network):
         network.lstm.save(fname)
     else:
         if args.strip:
-            network.clear_log()
+            print yellow('saving stripped network (without temporary variables)...')
```
zuphilip: The function `yellow` is defined here https://github.com/tmbdev/ocropy/blob/80f05404b5f0cbdbcfa60550dafcd8749685be65/ocrolib/common.py#L826 . Do we need that?

amitdo: No.

PR #145, opened 2016-11-06
Test page workflow and confidence measure, update coverage test

kba:
Requires coverage module: `pip2 install coverage`

Coverage reports should help finding untested/unused code (#143) and tests will be run on CI (PR upcoming).

It's not yet ready to merge but feedback and pointers on what to test is welcome.


run-test
```diff
@@ -1,13 +1,35 @@
...
-ocropus-hocr 'temp/????.bin.png' -o temp.html
-ocropus-visualize-results temp
-ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
+test_page() {
+    set -x
+    $RUNNER ocropus-nlbin tests/testpage.png -o temp
+    $RUNNER ocropus-gpageseg 'temp/????.bin.png'
+    $RUNNER ocropus-rpred -n 'temp/????/??????.bin.png'
+    $RUNNER ocropus-hocr 'temp/????.bin.png' -o temp.html
+    $RUNNER ocropus-visualize-results temp
```
zuphilip: Will this work in Circle CI? This needs AFAIK some graphical output device and it does not work when running as a docker container.

run-test
```diff
@@ -1,13 +1,35 @@
 #!/bin/bash -e
+# For coverage tests, run as
+#     coverage erase
+#     PATH=$PWD:$PATH RUNNER="coverage run --include=$PWD/**/*" ./run-test
+#     coverage html
+# See ./run-coverage
+RUNNER="${RUNNER:-python2}"
```
zuphilip: Can you add an explanation here about the idea of this variable?

run-test
```diff
@@ -1,13 +1,35 @@
 #!/bin/bash -e
+# For coverage tests, run as
+#     coverage erase
+#     PATH=$PWD:$PATH RUNNER="coverage run --include=$PWD/**/*" ./run-test
+#     coverage html
+# See ./run-coverage
```
zuphilip: Well, this lines are slightly different from the ones in `run-coverage` and I would argue that running tests is more frequent what users need than to run the coverage. Maybe explain these two use cases just shortly in two lines, e.g.
```bash
# For running the tests, use `./run-test`
# For coverage tests, use `./run-coverage` which will use these tests here
```

run-test
kba: This is the binary that the scripts are executed by. It defaults to `python2` if not set in the environment.

zuphilip: Okay, I suggest to put this into a comment line.

run-test
kba: I don't see why not but haven't tested it. With the `agg` backend, matplotlib should be runnable headless, can you open an issue for that in ocropy-docker?

run-test
kba: It's just an explanation on why you would want to set `RUNNER` variable in the environment. You also might want to use a different python version, e.g. `RUNNER=python3 ./run-test`.

zuphilip: Would something more simple like `RUNNER="coverage run --include=$PWD/**/*" ./run-test` without the lines before and afterwards work to demonstrate the RUNNER variable? The python3 example would also be nice to show.

run-test
zuphilip: Hm... now I just see a user warning:
```
root@0284af8aa1ad:/ocropy# ocropus-visualize-results temp
=== 0001
/usr/lib/python2.7/dist-packages/scipy/ndimage/interpolation.py:532: UserWarning
: From scipy 0.13.0, the output shape of zoom() is calculated with round() inste
ad of int() - for these inputs the size of the returned array has changed.
  "the returned array has changed.", UserWarning)
```
Maybe, I didn't remember correctly.

run-test
kba: Added an explanation. I wasn't aware of run-coverage when I started with this, so I added the commands here.

run-test
kba: That warning is unrelated but something to investigate. It's another rounding issue like #123

run-test
```diff
@@ -1,13 +1,38 @@
...
-ocropus-visualize-results temp
-ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
-
-echo "to see recognition results, type: firefox temp.html"
-echo "to see correction page, type: firefox temp-correction.html"
-echo "to see details on the recognition process, type: firefox temp/index.html"
+# 'RUNNER' is the binary that the scripts are executed by. It defaults to
+# 'python2' if not set in the environment.
+# Examples:
+#       PATH=$PWD:$PATH RUNNER='coverage run --include=$PWD/**/*' ./run-test
```
zuphilip: This should IMO be
```
#       PATH=$PWD:$PATH RUNNER='python2 -m coverage run --include=$PWD/**/*' ./run-test
```

run-test
```diff
@@ -1,13 +1,38 @@
...
+    echo "#     to see details on the recognition process, type: firefox temp/index.html"
+}
+
+test_conf() {
+    local TESTIMAGE=0079-01000d
+    mkdir -p temp
+    cp "$BASE/tests/$TESTIMAGE"* temp
+    set -x
+    $RUNNER ocropus-rpred temp/$TESTIMAGE.png
+    $RUNNER ocropus-errs  temp/$TESTIMAGE.gt.txt
```
zuphilip: I would suggest to use another copy of the `run-test` script for testing purposes only. The other use case is to provide some documentation how the different tools are called and working together (it is also mentioned in the `README.md`), and for this purpose these lines become IMO too confusing.

run-coverage
```diff
@@ -1,17 +1,10 @@
...
-python -m coverage run -p ocropus-rpred -n 'temp/????/??????.bin.png'
-python -m coverage run -p ocropus-hocr 'temp/????.bin.png' -o temp.html
-python -m coverage run -p ocropus-visualize-results temp
-python -m coverage run -p ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
-python -m coverage run -p ocropus-rpred tests/0079-01000d.png
-python -m coverage run -p ocropus-errs tests/0079-01000d.gt.txt
-python -m coverage run -p ocropus-econf tests/0079-01000d.gt.txt
-python -m coverage merge
+COVERAGE="python2 -m coverage"
+
```
zuphilip: Can we test that `coverage` is installed and otherwise exit with a user warning?

run-test
zuphilip: Then, we can also afterwards decide to switch Travis to the new testing script...

run-test
```diff
@@ -1,13 +1,35 @@
...
+    cp tests/$TESTIMAGE* temp
+    set -x
+    $RUNNER ocropus-rpred temp/$TESTIMAGE.png
+    $RUNNER ocropus-errs  temp/$TESTIMAGE.gt.txt
+    $RUNNER ocropus-econf temp/$TESTIMAGE.gt.txt
+}
+
+rm -rf temp
+test_page
+test_conf
```
zuphilip: (Also missing for future improvements: linegen, dewarp)

zuphilip:
@kba What are your plans for this PR? After solving the merge conflicts and addressing my comments, it seems for me okay to merge it.

run-test
zuphilip: Added

run-test
zuphilip: Done

run-coverage
zuphilip: I tried with
```sh
if !"$COVERAGE help"; then
   echo "You need to install the coverage module first, e.g. 'pip2 install coverage'"
   exit 1
fi
```
but somehow this does not work as expected... Any ideas?

kba:
```sh
if ! "$COVERAGE" help; then
   echo "You need to install the coverage module first, e.g. 'pip2 install coverage'"
   exit 1
fi
```

zuphilip:
Okay, that PR would be ready now from my perspective. @kba What do you think?

PR #146, opened 2016-11-07
lxml is required for ocropus-gtedit

kba:


zuphilip:
Looks good in general.

We have currently BeautifulSoup; do you know if this is used anywhere? Should we use both `lxml` and `BeautifulSoup` for the project here?

kba:
No, it would be better to use just one. lxml has a BS-like interface as well I think. @zuphilip Maybe open an issue for that?

PR #149, opened 2016-11-07
circle.yml file for Circle CI testing, run-test-ci

kba:
Not yet ready for merging, but if anyone's interested in trying out, here's a setup for continuously testing with CircleCI.

circle.yml:
* Set MPLBACKEND=agg to default to non-interactive matplotlib use

run-test simplified for reasonably fast testing
* rpred: only the first 15 lines
* hocr: ignore warnings about non-existing line texts

ocropus-rpred can run really slow (sometimes, up to 10 seconds per line) in the CI environment so I shortened the test to make CI faster. This could be also be copied to a CI-specific version of the test. See also #145


circle.yml
```diff
@@ -0,0 +1,18 @@
...
+  python:
+    version: 2.7.12
+  environment:
+    # Set matplotlb backend to the non-interactive antigrain image lib
+    MPLBACKEND: Agg
+dependencies:
+  pre:
+    # 'models' folder is cached, don't download twice
+    - cd models && wget -nc http://www.tmbdev.net/en-default.pyrnn.gz
+    # Pipe to cat to hide the progress bars
```
QuLogic: `pip -q`?

kba: `-q` hides INFO messages, including build errors which I do want to see. There is no option like `--no-progress-bar` to hide just the animated progress bar which messes with CircleCI web rendering of the console. Piping to cat hides the progress bar and keeps the error output.

BTW: I'm aware of #37 (thanks!) and reviewing your commits. Ideally, we'll have CircleCI for virtualenv and TravisCI for conda testing.

circle.yml
```diff
@@ -0,0 +1,18 @@
...
+    - pip install -r requirements_1.txt|cat
+    - pip install -r requirements_2.txt|cat
```
zuphilip: We have now only one list of requirements.

run-test
```diff
@@ -3,8 +3,8 @@
 rm -rf temp 
 ocropus-nlbin tests/testpage.png -o temp
 ocropus-gpageseg 'temp/????.bin.png'
-ocropus-rpred -n 'temp/????/??????.bin.png'
-ocropus-hocr 'temp/????.bin.png' -o temp.html
```
zuphilip: I suggest to change the `run-test` only in the PR #145. And I agree on

> This could be also be copied to a CI-specific version of the test. 

Yes, we should create a separate version of this script for CI.

zuphilip:
What is the status overall here? I am fine to add the `circle.yml` when it is updated and working. This enables us to test a different environment where ocropus can be installed, which would be great to have. Also integration still has to be activated.

kba:
I merged master and created a new file `run-test-ci` with the shorter line recognition tests for ci.

zuphilip:
Okay, this looks good. Is it ready for merging?

PR #150, opened 2016-11-07
ocrolib/h5utils.py: Move to OLD

kba:
`h5utils.py` is only imported by `OLD/ocropus-lattices` and
`OLD/ocropus-db`. It's the only tool outside `OLD` to use `tables`.
Hence moving h5utils.py to `OLD` allows dropping the requirement
on pytables and the two-step `requirements.txt` setup.


kba:
Passes `./run-test`: https://circleci.com/gh/kba/ocropy/60

amitdo:
What about` lattice.py` and `ngraphs.py`. Both aren't being used with the current lstm based engine.

tmbdev:
For new code, h5py is preferable anyway (old versions didn't support variable length arrays).

PR #151, opened 2016-11-07
Search for models relative to installed library.

kba:
> This works better than hard-coding /usr/local/share.

Extracted from #37

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -665,15 +665,11 @@ def expand_args(args):
     else:
         return args
 
-data_paths = [
+_data_paths = [
     ".",
-    "./models",
-    "./data",
-    "./gui",
-    "/usr/local/share/ocropus/models",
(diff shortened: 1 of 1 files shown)


ocrolib/common.py
```diff
@@ -684,13 +680,28 @@ def ocropus_find_file(fname,gz=1):
...
         if os.path.exists(full): return full
     if gz:
-        for path in data_paths:
-            full = path+"/"+fname+".gz"
+        for path in _data_paths:
+            full = os.path.join(path, fname + ".gz")
             if os.path.exists(full): return full
+
+    share_path = os.path.dirname(inspect.getfile(inspect.currentframe()))
+    share_path = os.path.join(share_path,
```
QuLogic: I think this maybe can be made a little nicer using `sysconfig`. Should I PR you or just open a new one here?

QuLogic: #152

PR #152, opened 2016-11-08
Search for models relative to installed library.

QuLogic:
This works better than hard-coding /usr/local/share.

Replaces #151 because I found a better way to find the `share` path.


ocrolib/common.py
```diff
@@ -665,34 +666,50 @@ def expand_args(args):
...
+        full = os.path.join(path, fname)
+        if os.path.exists(full):
+            return full
+    if gz:
+        for path in _data_paths:
+            full = os.path.join(path, fname + ".gz")
+            if os.path.exists(full):
+                return full
+
+    share_path = os.path.join(sysconfig.get_config_var("datarootdir"),
```
kba: Won't this depend on the location where Python was installed rather than where the module data was installed? For my box, this returns `/usr/share/ocropus` while `python setup.py install` installs the models to `/usr/local/share`. How about something like

```python
for prefix in ['.', '/usr/local/share/ocropus', '/usr/share/ocropus']:
  fpath = os.path.join(prefix, fname)
  if os.path.exists(fpath): return fpath
  if gz and os.path.exists(fpath+".gz"): return fpath + ".gz"
  for subdir in ["models", "data","gui"]:
    fpath = os.path.join(prefix, subdir, fname)
    if os.path.exists(fpath): return fpath
    if gz and os.path.exists(fpath+".gz"): return fpath + ".gz"

Comment (kba) [inline ocrolib/common.py]: Is the [`finddir`](https://github.com/tmbdev/ocropy/blob/master/ocrolib/common.py#L607) method used anywhere? Should the data-loading logic be moved to [`defaults.py`](https://github.com/tmbdev/ocropy/blob/master/ocrolib/default.py)?

Comment (QuLogic) [inline ocrolib/common.py]: Hmm, you know what, the original probably works better if we really want to tie it to the install location.

Your proposal doesn't really work in virtual environments by default (unless OCROPUS_DATA is set, which I'd like to avoid.)

It doesn't appear that `finddir` is used anywhere.

**Comment by QuLogic** [comment]
How is that then? Covers the other install locations as well.

File: ocrolib/common.py [hunk 2]
```diff
@@ -677,36 +678,34 @@ def expand_args(args):
...
+    possible_prefixes = [os.curdir]
+
+    local_share_path = os.path.dirname(inspect.getfile(inspect.currentframe()))
+    local_share_path = os.path.join(local_share_path,
+                                    os.pardir, os.pardir, os.pardir, os.pardir,
+                                    "share", "ocropus")
+    if os.path.isdir(local_share_path):
+        possible_prefixes += [local_share_path]
+
+    share_path = os.path.join(sysconfig.get_config_var("datarootdir"),
```
kba: If Python were installed in `/usr/local` wouldn't this be the same as `local_share_path`?

QuLogic: Yes, but I thought model installation was made optional. So this covers the case where someone has some system install _with_ models, and then they made a local install (e.g., `~/.local/lib/python`) _without_ models. Which might be what some developer would do if trying to save some room.

ocrolib/common.py
```diff
@@ -677,36 +678,34 @@ def expand_args(args):
...
-        if os.path.exists(full):
-            return full
-    if gz:
+    for prefix in possible_prefixes:
         for path in _data_paths:
-            full = os.path.join(share_path, path, fname + ".gz")
+            full = os.path.join(prefix, path, fname)
             if os.path.exists(full):
                 return full
+        if gz:
```
QuLogic: That's actually how it worked before, with the exception of the current directory, which used to check for both before the rest (and is now different, admittedly.)

I agree that it makes more sense to check for both file types in a single directory at a time.

ocrolib/common.py
kba: Sure, I see the point but if I run `sudo python.py install` with models outside, system-wide, models will be installed to `/usr/local/share/ocropus`. If I subsequently install ocropus to `~/.local` and run it from there, it will look in `~/.local/share/ocropus` (because of `local_share_path`) and `/usr/share/ocropus` (because of `share_path` and Python being installed to `/usr/share`) but not in `/usr/local/share/ocropus`.

ocrolib/common.py
kba: it was more a general observation, not something introduced by you. I've wondered about that before, maybe there was a reasoning behind it.

QuLogic:
I made it check in something like `/usr/local` as well, and tried to remove duplicate search paths.

zuphilip:
@kba Can you look at the newest changes here?

kba:
Merged with current master, will check later.

ocrolib/common.py
```diff
@@ -680,16 +680,23 @@ def ocropus_find_file(fname, gz=True):
...
-                                    "share", "ocropus")
-    if os.path.isdir(local_share_path):
+    install_share_path = os.path.dirname(
+        inspect.getfile(inspect.currentframe()))
+    install_share_path = os.path.join(install_share_path, os.pardir, os.pardir,
+                                      os.pardir, os.pardir, "share", "ocropus")
+    if os.path.isdir(install_share_path):
+        possible_prefixes += [install_share_path]
+
+    local_share_path = os.path.join(sysconfig.get_config_var("prefix"),
```
kba: Won't this result in `/usr/local/local/share` if python is installed with `prefix=/usr/local`? Why not hard-wire `/usr/local/share/ocropus`? Or use `os.path.join(sysconfig.get_config_var('data'), 'share', 'ocropus')` though I'm not sure what this is set to if Python isn't installed with `prefix=/usr`?

kba:
@QuLogic Can you take a look at https://github.com/kba/ocropy/commit/4907866f97a2e14db892a47717e32bca62d10bac? Your algorithm, but normalizing the install_share_path, hard-wiring `/usr/local/share/ocropus`, uniq-ing the and isdir-checking the entries when searching, plus documentation.

ocrolib/common.py
QuLogic: Hmm, you might be right. I don't have an install like that to check though.

QuLogic:
The problem with `np.unique` is that it does not preserve order.

kba:
> The problem with np.unique is that it does not preserve order.

Was not aware of that, thanks.

kba:
Fixed with preserved order in https://github.com/kba/ocropy/commit/a6630be78c02dcab6967f65075fd85dbb77bb33b

QuLogic:
It seems to make sense; not sure we need to hard-code /usr/local/share, but it'll be skipped if missing, so I guess it doesn't hurt.

kba:
Thanks for the review @QuLogic I fixed those now. Having `/usr/local/share/ocropus` hard-wired is probably overkill but since it's referenced as such in a few places, it should ensure backwards-compatibility.

Tests seem to work, so I'll push the changes here. Unless someone objects I'll merge this tomorrow.

zuphilip:
Thank you both for the work and discussion here!

PR #154, opened 2016-11-15
Cleanup imports in common.py, lstm.py, extract exceptions

kba:
* Move exceptions to its own module
* Make all imports explicit, so static code analysis tools like pyflakes can help
* remove redundant deprecated decorators
* Remove old load_components
* Remove old draw_pseg/draw_aligned

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -824,35 +824,6 @@ def save_component(file,object,verbose=0,verify=0):
         with open(file,"rb") as stream:
             pickle.load(stream)
 
-def load_component(file):
-    """Load a component. This handles various special cases,
-    including old-style C++ recognizers (soon to be gotten rid of),
-    python expressions ("=package.ObjectName(arg1,arg2)"),
-    and simple pickled Python objects (default)."""
-    if file[0]=="=":
-        return pyconstruct(file[1:])
-    elif file[0]=="@":
-        file = file[1:]
-    with open(file,"r") as stream:
-        # FIXME -- get rid of this eventually
-        start = stream.read(128)
-    if start.startswith("<object>\nlinerec\n"):
-        # FIXME -- get rid of this eventually
-        warnings.warn("loading old-style linerec: %s"%file)
-        result = RecognizeLine()
-        import ocropus
-        result.comp = ocropus.load_IRecognizeLine(file)
-        return result
-    if start.startswith("<object>"):
-        # FIXME -- get rid of this eventually
-        warnings.warn("loading old-style cmodel: %s"%file)
-        import ocroold
-        result = ocroold.Model()
-        import ocropus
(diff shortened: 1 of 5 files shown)


ocrolib/common.py
```diff
@@ -11,97 +11,32 @@
 import warnings
 import inspect
 import glob
+import pickle
 import cPickle
+from ocrolib.exceptions import (BadClassLabel, BadInput, FileNotFound,
```
QuLogic: Should move down to third block (first-party imports).

ocrolib/common.py
```diff
@@ -11,97 +11,32 @@
...
+                                OcropusException)
 
 import numpy
-from numpy import *
+from numpy import (amax, amin, array, bitwise_and, clip, dtype, mean, minimum,
+                   nan, ndarray, sin, sqrt, zeros)
 import pylab
-from pylab import imshow
-from scipy.ndimage import morphology,measurements
+from pylab import (clf, cm, ginput, gray, imshow, ion, subplot, where)
```
QuLogic: I've got a branch that removes `pylab` imports; should I open a PR for that?

ocrolib/common.py
```diff
@@ -11,97 +11,32 @@
...
 import glob
+import pickle
 import cPickle
+from ocrolib.exceptions import (BadClassLabel, BadInput, FileNotFound,
+                                OcropusException)
 
 import numpy
-from numpy import *
+from numpy import (amax, amin, array, bitwise_and, clip, dtype, mean, minimum,
+                   nan, ndarray, sin, sqrt, zeros)
```
QuLogic: The branch also standardizes these imports too...

ocrolib/common.py
```diff
@@ -11,97 +11,32 @@
...
 from default import getlocal
-from toplevel import *
+from toplevel import (checks, ABINARY2, AINT2, AINT3, BOOL, DARKSEG, GRAYSCALE,
+                      LIGHTSEG, LINESEG, PAGESEG)
 import chars
 import codecs
 import ligatures
 import lstm
 import morph
 import multiprocessing
```
QuLogic: Should move up to first block (standard library imports).

zuphilip:
I see that the exceptions are deleted in `ocrolib/common.py` but I don't see another module with them. Is that (new) file missing here?

ocrolib/common.py
kba: You mean https://github.com/QuLogic/ocropy/commits/standard-import? Sure, open a PR with the pertinent commits, looks really helpful, I'll happily merge/cherry-pick.

QuLogic: Yes, that's the one; though I just noticed that it's based on the Python 3 work, so it'll be a pain to cherry-pick. Should get that in ASAP and then the rest would be easier.

PR #155, opened 2016-11-15
py3k: Use print function instead of statement

QuLogic:
I split this out of #114 for easier review.

There are some changes to OLD because they were there originally, but I assume there's no need to really update the rest of them. Though I wonder why you don't just delete those files? That's why you've got git history...

### DIFF
--- a/OLD/ocrolib/ngraphs.py
+++ b/OLD/ocrolib/ngraphs.py
@@ -1,6 +1,8 @@
 # TODO:
 # - handle UTF-8 inputs correctly
 
+from __future__ import print_function
+
 from pylab import *
 from collections import Counter,defaultdict
 import re
@@ -47,7 +49,7 @@ def safe_readlines(stream,nonl=0):
         try:
             line = stream.readline()
         except UnicodeDecodeError as e:
-            if not once: print lineno,":",e
+            if not once: print(lineno, ":", e)
             once = 1
             return
         if line is None: return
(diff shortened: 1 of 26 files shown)


kba:
:+1: Looks good in general, will test later.

> There are some changes to OLD because they were there originally, but I assume there's no need to really update the rest of them. Though I wonder why you don't just delete those files? That's why you've got git history...

Does not really hurt to have an extra folder with source code, makes it easier to search and understand dead code and it can be useful to dig around in there to understand the development. For me documentation, cleaning up working code and testing is more important, deleting OLD is up to @tmbdev. I wouldn't fix code in OLD though.

OLD/ocrolib/ngraphs.py
```diff
@@ -106,11 +108,11 @@ def __init__(self,*args,**kw):
         NGraphsCounts.__init__(self,*args,**kw)
     def buildFromFiles(self,fnames,n):
         """Given a set of files, build the log posteriors."""
-        print "reading",len(fnames),"files"
+        print("reading", len(fnames), "files")
         counter = self.computeNGraphs(fnames,n)
-        print "got",sum(counter.values()),"%d-graphs"%(n,)
+        print("got", sum(counter.values()), "%d-graphs" % (n, ))
```
zuphilip: This looks strange: After the last comma there is nothing but the closing parenthesis.

zuphilip:
Thank you very much @QuLogic ! Sorry for the delay.

PR #157, opened 2016-11-16
py3k: Use new-style exceptions

kba:
From #114 

* Do not raise strings but proper exceptions
* Use new-style except syntax

Comments on possible imrpovements: https://github.com/tmbdev/ocropy/pull/114#pullrequestreview-3418437


QuLogic:
This and #155 will conflict with each other.

kba:
Yes, but conflicts easily fixed.

zuphilip:
In another PR we also deal with exceptions: https://github.com/tmbdev/ocropy/pull/154/files#diff-0a7898a0c1927c32daf42e51d098ba7c . Any reason we do the exceptions here and there differently?

QuLogic:
That PR is about refactoring where the custom exceptions are; this one is about fixing the syntax for Python 3.

zuphilip:
Superseeded by #175

Issue #112, opened 2016-10-04
ocropus-hocr should use property `image` instead of `file`


zuphilip:
In the hocr fileformat it is important to link to the image from each `ocr_page`, see http://kba.github.io/hocr-spec/1.2/#image,  and this should look like

``` html
<div class='ocr_page' title='image 433934212_0017.bin.png'>
```

where currently ocropus produces the nonstandard

``` html
<div class='ocr_page' title='file 433934212_0017.bin.png'>
```

I guess it is enough to change this line https://github.com/tmbdev/ocropy/blob/master/ocropus-hocr#L66

Issue #107, opened 2016-08-12
Coordinates in bbox must be from top to bottom


zuphilip:
From the [hocr specs](https://github.com/kba/hocr-spec/blob/master/hocr-spec.md#5-typesetting-related-elements):

> the bounding box of the page; for pages, the top left corner must be at (0,0), so a typical page bounding box will look like bbox 0 0 2300 3200

However, the output of `ocropy-hocr` will produce `bbox` with coordinates measured w.r.t. to the bottom left corner. See for example [book.html.txt](https://github.com/tmbdev/ocropy/files/414890/book.html.txt) where the lines in the image are from top to bottom but the the y-coordinates are decreasing in the hocr output.

Maybe it is enough to replace https://github.com/tmbdev/ocropy/blob/master/ocropus-hocr#L91 with

``` python
 y0,x0,y1,x1 = regions.bbox(i)
```

See also https://github.com/tmbdev/ocropy/blob/master/ocrolib/common.py#L427-L438

PR #160, opened 2016-12-02
Made output more compatible to the hOCR spec.

jze:
- properties "image" and "bbox" for the "ocr_page" element.
- correct orientation of coordinate system


zuphilip:
Wow, that looks good! 🎉  This should solve the issues #112 and #107. Thank you very much! I will have to look closer some time later. Stay tuned.

zuphilip:
Thank you very much @jze!

PR #165, opened 2016-12-30
Print characters instead of repr

ChillarAnand:


ocropus-rtrain
```diff
@@ -296,10 +296,10 @@ for trial in range(start,args.ntrain):
     acs = lstm.translate_back(network.aligned)
     gta = "".join(codec.decode(acs))
     if not args.quiet:
-        print "%d %.2f %s"%(trial,network.error,line.shape),fname
-        print "   TRU:",repr(transcript)
-        print "   ALN:",repr(gta[:len(transcript)+5])
-        print "   OUT:",repr(pred[:len(transcript)+5])
+        print("%d %.2f %s %s"%(trial, network.error, line.shape, fname))
+        print("   TRU:{}".format(transcript.encode('utf-8')))
```
QuLogic: You shouldn't encode the input to `print`; the console encoding is not necessarily utf-8.

Issue #132, opened 2016-10-30
Writing on sys.stderr


zuphilip:
There is quite a bunch of statements `sys.stderr.write` where some information is written on standard error stream instead of normal standard output stream, cf. https://github.com/tmbdev/ocropy/search?utf8=%E2%9C%93&q=sys.stderr.write

kba:
Is that a problem per se? Making logging consistent is a good thing but writing warnings/errors to STDERR can be useful as well.

tmbdev:
Yes; if you want both stderr and stdout together, just run the command with 

    command ... args ... 2>&1

zuphilip:
It is not a general problem, but I had the impression that maybe some of the print statements would be more consistent to use standard output instead, e.g. the information about the detected errors in [ocropus-econf](https://github.com/tmbdev/ocropy/blob/97ace5581b4c4deb95cfcd17b595b0ff1da36c25/ocropus-econf#L87-L97) is reported with these lines
```py
sys.stderr.write("errors    %8d\n"%errs)
sys.stderr.write("missing   %8d\n"%missing)
sys.stderr.write("total     %8d\n"%total)
sys.stderr.write("err       %8.3f %%\n"%(errs*100.0/total))
sys.stderr.write("errnomiss %8.3f %%\n"%((errs-missing)*100.0/total))

if args.confusion>0:
    for (a,b),v in counts.most_common(args.confusion):
        print "%d\t%s\t%s"%(v,a,b)

print errs*1.0/total
```
Thus, we have some `sys.stderr.write` and some normal `print` statements. Shouldn't we change some of these statements? (Please note that such changes in this file will currently interfere with https://github.com/tmbdev/ocropy/pull/131)

stweil:
Using stderr for error messages is fine. Mixing stdout and stderr for normal program output results in several problems:

* There is no synchronization between stdout and stderr, so messages appear in an unexpected order (also if using `2>&1`). Typically stdout is buffered while stderr is unbuffered. For example `ocropus-err` outputs file results, then the summary followed by more file results while the summary is expected at the end.

* Users don't expect normal output on stderr. If they use something like `ocropus-err ... | less`, they will see output, but miss the summary.

Philipp, either apply PR #131 first (then I'll rebase #170), or apply #170 (then #131 can be rebased, either by you or by me).

PR #170, opened 2017-01-07
Print summaries to stdout instead of stderr 

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/ocropus-econf
+++ b/ocropus-econf
@@ -8,7 +8,7 @@ from pylab import *
 from ocrolib import edist
 
 # disable rank warnings from polyfit
-warnings.simplefilter('ignore',numpy.RankWarning) 
+warnings.simplefilter('ignore',numpy.RankWarning)
 
 parser = argparse.ArgumentParser(description = """
 Compute the edit distances between ground truth and recognizer output.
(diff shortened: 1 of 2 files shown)


zuphilip:
Thank you very much! It looks that this will solve #132 in these two files.

ocropus-econf
```diff
@@ -84,11 +84,11 @@ for fname,e,t,m,c in outputs:
 if perfile is not None: perfile.close()
 if allconf is not None: allconf.close()
 
-sys.stderr.write("errors    %8d\n"%errs)
-sys.stderr.write("missing   %8d\n"%missing)
-sys.stderr.write("total     %8d\n"%total)
-sys.stderr.write("err       %8.3f %%\n"%(errs*100.0/total))
-sys.stderr.write("errnomiss %8.3f %%\n"%((errs-missing)*100.0/total))
+print "errors    %8d"%errs
```
QuLogic: Use `print()` and be Python 3 ready. There's only one argument, so there's no effect on Python 2.

stweil: Good hint. I'll add a commit which fixes that tomorrow.

amitdo:
Stefan, please add `from __future__ import print_function`

ocropus-econf
```diff
@@ -84,14 +84,14 @@ for fname,e,t,m,c in outputs:
 if perfile is not None: perfile.close()
 if allconf is not None: allconf.close()
 
-print "errors    %8d"%errs
-print "missing   %8d"%missing
-print "total     %8d"%total
-print "err       %8.3f %%"%(errs*100.0/total)
-print "errnomiss %8.3f %%"%((errs-missing)*100.0/total)
+print("errors    %8d"%errs)
```
amitdo: Stefan, please add from __future__ import print_function on top

stweil: Is it also needed if print only gets one argument? At least for Python 2.7.9 it works without that statement.

zuphilip: Maybe not needed here, but we will add that statement anyway during PR #155 see https://github.com/tmbdev/ocropy/pull/155/files#diff-672c98336ceb05ab25dc90ed9934175f (I hope we can continue on these other PR the next weeks)

zuphilip: Thus, IMO we can add this statement also afterwards when we deal with the other occurrences of printing statements.

amitdo: I thought it will fail in 2.6/2.7... but you reminded me that in this case it's fine.

zuphilip:
@stweil FYI I just tried out the solve merge conflicts here on GitHub and it worked well.

zuphilip:
Thank you very much @stweil !

Issue #171, opened 2017-01-10
ocropus-gpageseg: Too many column separators detected


lehzwo:
## Current Behavior
When passing _n_ as a value for the maximum number of white column separators via `--maxcolseps`, _n+1_ white column separators are detected instead. Therefore it is not possible to process an image without taking at least one white column separator into account.
The detection of black column separators via `-b` on the other hand detects the correct number of black separators if the passed value of `--maxseps ` does not equal 0. In this case all black separators are detected.

## Possible Solution

At _line 212_ in _compute_colseps_conv_ in _ocropus-gpageseg_
`seps = morph.select_regions(seps,sl.dim0,min=args.csminheight*scale,nbest=args.maxcolseps+1)`,
`args.maxcolseps+1` is passed instead of `args.maxcolseps`. Therefore an additional whitespace separator is selected.

When changing that line to 
`seps = morph.select_regions(seps,sl.dim0,min=args.csminheight*scale,nbest=args.maxcolseps)`,the whitespace separator detection behaves like the black separator detection mentioned above (if 0 is passed, all separators are detected). This can be prevented by adding the check `if nbest > 0:` to the block 
```
for i in best[-nbest:]:
   if scores[i]<=min: continue
   keep[i+1] = 1
```
at line 198 in _select_regions_ in _morph.py_

zuphilip:
@lehzwo Thank you for this report and I have also seen your PR. I will need more time to look at it, think about and try out some things... 

Do you have an example image which shows this behavior and how your PR solves the issue? I am not sure but I have something in mind, that the a white column separator is also needed for the "left border" and therefore even in a single-column text. This might be related to https://github.com/tmbdev/ocropy/issues/115.

The change might be critical for anyone who relies on the "old behavior". Are there alternatives which are extending the old behavior with the new functionality?

Let me also ping @jze and @1a1a11a which worked on tables in ocropy before. Any reactions from you?

lehzwo:
I processed the image [http://dibiki.ub.uni-kiel.de/viewer/fullscreen/PPN828360804/32/](url) via _ocropus-nlbin_ without additional parameters to get the binary image. Without the changes of the PR, the following command
`ocropus-gpageseg -n -d --maxcolseps 0`
detects one white column separator in the middle of the image. This separator is taken into account when creating the line seeds.

[...]

PR #172, opened 2017-01-10
Detect the passed number of separators

lehzwo:
Implementation of **Possible Solution** specified in issue #171

### DIFF
--- a/ocrolib/morph.py
+++ b/ocrolib/morph.py
@@ -195,9 +195,10 @@ def select_regions(binary,f,min=0,nbest=100000):
     scores = [f(o) for o in objects]
     best = argsort(scores)
     keep = zeros(len(objects)+1,'i')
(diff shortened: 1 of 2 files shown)


zuphilip:
@lehzwo Okay, I check your solution more carefully and it looks good for me. 👍  We should maybe just increase the default value of this parameter to `3` to be more conform with the "old behaviour".

Moreover, it would then be possible to delete the parameter `-b` completely and just start with a default value of `0` for `--maxseps` but maybe we skip this to an issue first for discussion.

The merge conflict should be easy to solve, but I can also do this. Let me know what is best for you.

lehzwo:
@zuphilip The change of the default parameter as well as the potential change of the behaviour of the `-b` parameter sound good to me.

zuphilip:
@lehzwo Thank you very much for this PR!

PR #179, opened 2017-01-31
Update README.md

zuphilip:
Please note also that Travis CI is now activated. CC @kba


zuphilip:
Thank you @jze and @lunactic for the review/testing!

PR #184, opened 2017-02-11
Add option `--naming` for different line name patterns

zuphilip:


wrznr:
This a very valuable enhancement. However, I would drop the predefined choices and rely on pure format strings only.

zuphilip:
@wrznr Do you mean that with the `naming` option only a user-defined pattern can be indicated (deleting the `hex` and `dec` cases)? Then, this parameter should be optional and by default the hexadecimal pattern be used as it is currently (to not break any possible dependencies on tools based on ocropus).

Then, these calls would be possible:
```sh
ocropus-gpageseg test.bin.png
# by default the pattern "01%04x" is used

ocropus-gpageseg test.bin.png --naming %04d
# pattern with 4 decimal digits
```

zuphilip:
There may be more places to attribute different length of file names:
```
root@zuphilip-VirtualBox:/etc/ocropy# grep -r \?\?\?\?\?\? *
ocrolib/common.py:        return sorted(glob.glob(args[0]+"/????/??????.png"))
ocropus-hocr:xhfiles = python.sum([glob.glob(d+"/??????.xheight") for d in dirs],[])
ocropus-hocr:    lfiles = python.sum([glob.glob(d+"/??????.bin.png") for d in dirs],[])
ocropus-visualize-results:            images = sorted(glob.glob("??????.bin.png"))
ocropus-visualize-results:        for fname in sorted(glob.glob(d+"/??????.txt")):
README.md:    ./ocropus-rpred -Q 4 -m models/fraktur.pyrnn.gz 'book/????/??????.bin.png'
run-coverage:python -m coverage run -p ocropus-rpred -n 'temp/????/??????.bin.png'
run-coverage:python -m coverage run -p ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
run-test:ocropus-rpred -n 'temp/????/??????.bin.png'
run-test:ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
run-test-ci:ocropus-gtedit html temp/????/??????.bin.png -o temp-correction.html
```

amitdo:
https://www.mail-archive.com/ocropus@googlegroups.com/msg00807.html
https://www.mail-archive.com/ocropus@googlegroups.com/msg00850.html

zuphilip:
@amitdo What do you want to point me to with these two links?

amitdo:
In some point in history, Tom switched from `%04d` to` %04x.`

Just digging in ocropus history...

PR #190, opened 2017-03-09
Adjust reading order when there are no columns

zuphilip:
When there are no white column separators and no black
columns separators, then we determine the reading order
by simply looking which line is above which other lines.

### DIFF
--- a/ocrolib/psegutils.py
+++ b/ocrolib/psegutils.py
@@ -121,7 +121,7 @@ def extract_masked(image,linedesc,pad=5,expand=0):
     line = where(mask,line,amax(line))
     return line
 
-def reading_order(lines,highlight=None,debug=0):
+def reading_order(lines,nocolumns=False,highlight=None,debug=0):
     """Given the list of lines (a list of 2D slices), computes
     the partial reading order.  The output is a binary 2D array
     such that order[i,j] is true if line i comes before line j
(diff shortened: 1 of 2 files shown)


zuphilip:
The ordering of the textual parts on the same line is sometimes wrong.

mittagessen:
This might not be entirely pertinent to this pull request but inclusion of an equivalent switch arose a few days ago in kraken (from an RTL reading order bug). I am not confident that there aren't edge cases where horizontal intra-column ordering is desirable that such a switch would disable. Have you looked into scenarios where such a switch would fail?

It isn't like the current segmenter is particulary good and I still plan on replacing it reasonably soon but it still shouldn't be made worse for some inputs.

zuphilip:
@mittagessen Another PR #118 is also about the reading order. The picture there shows some of the cases which you might want to test as edge cases. The current code about this is hard to understand and will also only provide a partial ordering which is then extended to the transient hull later.

Maybe the question is also: do we expect that the columns recognition are not giving us all the details which we then have to handle in the reading order step?

mittagessen:
I would expect the reading order determination to be completely independent of column detection, just as it is right now, mainly because the metric for column separation is different from the one used to separate lines horizontally. 

Disabling the horizontal ordering results in minor vertical variations in typesetting causing reordering. A theoretical example I found is certain (single column) poetry that is justified through whitespace in the middle of the line; without y_overlap logic the ordering of the two line parts should be mostly random.

Admittedly, that's an esoteric use case but the proposed behavior would be worse for these inputs than the old one.

Issue #79, opened 2016-01-16
Possible bugs?


crazylyf:
Hello,
When I running ocropus-ltrain, it will occasionally warning: "FloatingPointError: overflow encountered in exp", and the program seems to restart from the nearest saved state. The problem occurs mainly in the "ffunc" function in lstm.py, which defines the softmax function using: 1.0/(1.0+exp(-x)). Same problem also occurs in the "sigmoid" function. I think this may be caused by large values in x. In the CLSTM source code, the values x is clipped to 20 for positive values, and -20 for negtive values. After clipping like this, the program goes well without warning.

Another problem is that the "backward" method in class "Parallel" returns None. This is correct for 1-layer BLSTM system, but for multiple layers BLSTM configuration which stacking paralleled BLSTM one over another, this will lead to error, as the deltas of subsequent layer is assigned as the current deltas. So, maybe the method should return deltas. 

Best,

tmbdev:
I'll keep the bug open, but for LSTM training, at this point, you're probably better off using the CLSTM implementation. It is almost a drop-in replacement and considerably faster.

If you want to submit patches for a "safe" sigmoid and for the BLSTM, please do.

zuphilip:
Fixed in #201.

zuphilip:
Second part may still be open:

> Another problem is that the "backward" method in class "Parallel" returns None. This is correct for 1-layer BLSTM system, but for multiple layers BLSTM configuration which stacking paralleled BLSTM one over another, this will lead to error, as the deltas of subsequent layer is assigned as the current deltas. So, maybe the method should return deltas.

Issue #49, opened 2015-07-20
FloatingPointError while training


lesliekim:
I followed the "run-rtrain" document to train RNN. I got the training set by " tar -zxf tests/uw3-500.tgz".  I didn't change anything, neither RNN structure nor parameters. I meet the following problem while training: FloatingPointError: overflow encountered in exp. However, the program does not stop. It keeps iterating while showing the above FloatingPointError message again and again. 
I reduce the learning rate to half of the default value. But the same problem occurs. Do I train the network in the wrong way?

tmbdev:
The default parameters should work without problems. Can you say more about your environment (Python version, OS, compiler version, etc.)?

You might want to give CLSTM a try as well; it's the same algorithm in C++.

lesliekim:
My environment: python 2.7.6  
                           Ubuntu 14.04.1 LTS \n \l
                           gcc version 4.8.4

fafo:
I'm getting the same errors. I've tried clipping the values, but the model seems keeping very high error rates (I'm experimenting with training samples taken from a mobile camera, since we want to recognize receipts uploaded with mobiles)

def ffunc(x):
    "Nonlinearity used for gates."
    # cliping to avoid overflows
    return 1.0/(1.0+exp(clip(-x, None, 700)))

tmbdev:
Sorry for not responding earlier. Can you provide a backtrace?

You might also add a line of python to the main program to turn overflows into errors (when I get around to it, I may do this myself).

zuphilip:
Should be fixed with #201.

Issue #5, opened 2014-11-13
error while training


stexandev:
After executing  (on 156 files of groundtruth text and imagery):
  ocropus-rtrain gt/????/*.png -F 10000 -o mub_combined &
I've got the following reproduceable error: 

454 150.32 (1486, 48) gt/0001/01000b.bin.png
   TRU: u'quod dicitur Fulda, quod est situm in pago Grapfeld, constructum in honore sancti'
   ALN: u'quuod dicituur Fuulda, qquod et situumm in pagoo Grapfeld, construuctuuumm in honnore '
   OUT: u' iiii ii te ti imm tm e iii eutmut m mi eii '
# oops, got FloatingPointError overflow encountered in exp
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-rtrain", line 228, in <module>
    pcs = network.trainSequence(line,cs,update=do_update,key=fname)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 863, in trainSequence
    self.outputs = array(self.lstm.forward(xs))
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 587, in forward
    xs = net.forward(xs)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 636, in forward
    outputs = [net.forward(xs) for net in self.nets]
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 545, in forward
    self.WIP,self.WFP,self.WOP)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 419, in forward_py
    go[t] = ffunc(gox[t])
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 367, in ffunc
    return 1.0/(1.0+exp(-x))
FloatingPointError: overflow encountered in exp
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-rtrain", line 232, in <module>
    network = ocrolib.load_object(last_save)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/common.py", line 502, in load_object
    fname = ocropus_find_file(fname)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/common.py", line 680, in ocropus_find_file
    if os.path.exists(fname):
  File "/usr/lib/python2.7/genericpath.py", line 18, in exists
    os.stat(path)
[...]

PR #201, opened 2017-04-18
Clip exponential in ffunc to avoid overflow

zuphilip:
This should avoid (hopefully) some possible FloatingPointError overflow errors.

The sigmoid function ffunc is for any x<-20 and x>20 already 0 resp. 1 up to 10^-9
and cutting will therefore not change the function substantially.

This idea is from @tmbdev in https://github.com/tmbdev/ocropy/issues/5#issuecomment-65329170
Implemented first in https://github.com/tmbdev/ocropy/issues/49#issuecomment-139919235
Additional infos from https://github.com/tmbdev/ocropy/issues/79#issue-126999594

### DIFF
--- a/ocrolib/lstm.py
+++ b/ocrolib/lstm.py
@@ -380,7 +380,8 @@ class MLP(Network):
 
 def ffunc(x):
     "Nonlinearity used for gates."
-    return 1.0/(1.0+exp(-x))
+    # cliping to avoid overflows
+    return 1.0/(1.0+exp(clip(-x,-20,20)))
 def fprime(x,y=None):
     "Derivative of nonlinearity used for gates."
     if y is None: y = sigmoid(x)


zuphilip:
Note [that some lines above](https://github.com/tmbdev/ocropy/blob/66c6007a218c44845bdd9048228f408d337dd717/ocrolib/lstm.py#L67-L71) we have:
```py
def sigmoid(x):
    """Compute the sigmoid function.
    We don't bother with clipping the input value because IEEE floating
    point behaves reasonably with this function even for infinities."""
return 1.0/(1.0+exp(-x))
```

* Do we need both identical functions `ffunc` and `sigmoid`?
* What are "IEEE floating points"?
* Should we reconsider the comment here?

mittagessen:
-20/20 seems to be overly conservative but should work. Ultimately the gradients are important and they are negligible even at -10/10. On another note ``scipy.special.expit`` is significantly faster than `.0/(1.0+exp(-x))`, so you might consider using that. 

You won't need both `ffunc` and `sigmoid` as pickling only retains references to classes and function. Adjusting the forward/backward methods of the appropriate classes should be enough.

``IEEE floating points`` are standard IEEE 754 floating point numbers. The comment suggests they are stable even for extreme values. The linked bug reports seem to contradict that notion.

kba:
I'd like to merge it, and/or try @mittagessen's proposed scipy fix. @zuphilip do you have data to test this?

zuphilip:
@kba I think it is good to merge this. Maybe reformulate the contradicting comment before... I don't have data for reproduce the error, but there is at least one confirmation https://github.com/tmbdev/ocropy/issues/246#issuecomment-331435380 that after this fix it worked fine.

kba:
Yes, this will fix a breaking bug, so let's merge it. Just would be interesting to test `scipy.special.expit`. But working is better than fast.

zuphilip:
@mittagessen and @kba Thank you for helping here!

Issue #176, opened 2017-01-23
Standardize imports


zuphilip:
From https://github.com/tmbdev/ocropy/pull/154#discussion_r87943472 by @QuLogic :

> I've got a branch that removes pylab imports; should I open a PR for that?

> You mean https://github.com/QuLogic/ocropy/commits/standard-import? Sure, open a PR with the pertinent commits, looks really helpful, I'll happily merge/cherry-pick.

> Yes, that's the one; though I just noticed that it's based on the Python 3 work, so it'll be a pain to cherry-pick. Should get that in ASAP and then the rest would be easier.

Maybe it is also now easier to redo the work which was done in the PRs, because the forks may be too different now.

PR #206, opened 2017-04-28
Standardize Imports I, #176

zuphilip:
The import statements for the main commands ocropus-* are
standardized:
 * Put different import statements into different lines
 * Group them into (1) standard imports, (2) third party
   imports, (3) local modules
 * Avoid the "from pylab import *"
 * Use rather "import numpy as np", "import matplotlib.pyplot as plt"

Moreover, some unused import statements were deleted and a very
few nits for readibility are done.

This work is heavily based on the work by @QuLogic in the branch
https://github.com/QuLogic/ocropy/commits/standard-import
See also #176 .

All commands from the current pipeline are tested to still work
in the usual calls (especially run-test and run-rtrain). Even more, the
outputs before and afterwards is exactly the same.


zuphilip:
I tried to automatically check the addition of a namespace to the relevant functions (i.e. these are the ones I found so far in the ocropus code) with these lines:
```
# used functions from numpy need now a prefix 'np.'
perl -ne 'print "$. $_" if /([^.]|^)(amax|amin|arange|array|asarray|clip|dtype\(|isnan|linspace|maximum\(|mean|median\(|minimum\(|ones|random|rand|randn|randint|transpose|var|vstack|zeros)/' ocropus-*

# used functions from matplotlib.pyplot need now a prefix 'plt.'
perl -ne 'print "$. $_" if /([^.]|^)(clf|cm.gray|cm.hot|draw|figure|gca|ginput|gray\(|imshow|\bion\(|\bplot|rc\(|rcParams|savefig|subplot|title)/' ocropus-*
```

ChillarAnand:
Looks good. Any updates on this? We can also use importmagic to automatically fix imports.

ocropus-dewarp
```diff
@@ -47,8 +43,8 @@ for fname in inputs:
     base,_ = ocrolib.allsplitext(fname)
     try:
         line = ocrolib.read_image_gray(fname)
-        lnorm.measure(amax(line)-line)
-        line = lnorm.normalize(line,cval=amax(line))
+        lnorm.measure(np.amax(line)-line)
+        line = lnorm.normalize(line,cval=np.amax(line))
         scipy.misc.imsave(base+".dew.png",line)
```
zuphilip: We still need `scipy` here, cf. Travis output.

ocropus-linegen
```diff
@@ -114,36 +122,41 @@ assert len(lines)>0
...
 
+
 def rgeometry(image,eps=0.03,delta=0.3):
-    m = array([[1+eps*randn(),0.0],[eps*randn(),1.0+eps*randn()]])
+    m = np.array([[1+eps*np.random.randn(),0.0],[eps*np.random.randn(),1.0+eps*np.random.randn()]])
     w,h = image.shape
-    c = array([w/2.0,h/2])
-    d = c-dot(m,c)+array([randn()*delta,randn()*delta])
+    c = np.array([w/2.0,h/2])
+    d = c-dot(m,c)+np.array([np.random.randn()*delta,np.random.randn()*delta])
```
zuphilip: `NameError: global name 'dot' is not defined` -> Thus, I guess this should be `np.dot(...)`

zuphilip:
@ChillarAnand Any checking with `importmagic` would be appreciated. Moreover, I would like to wait for #234.

ChillarAnand:
importmagic doesnt work out of the box. I have written a wrapper script and it works only on Python 3. 

Any timeline for [this PR](https://github.com/tmbdev/ocropy/pull/114)?

zuphilip:
Okay, I understand. Actually, we still concentrate on Python 2 and the PR you mentioned is for a separate Python 3 branch. In general it might be hard work to have the whole OCRopus work for Python 2 and Python 3.

zuphilip:
Wow, that was fast :-) I hope that no error was created in the last round of solving merge conflict, because I didn't test it myself much, but simply relied on Travis and the extended tests we have now...

kba:
:shipit: :shipit: :shipit: 

But seriously, if we'll see regressions from this, it will be fairly obvious, missing imports or undeclared vars.

PR #209, opened 2017-04-29
Unit tests

zuphilip:
Make sure that unit tests are tested during CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,4 +45,5 @@ install:
 script:
   - mkdir ../test_folder
   - cd ../test_folder
+  - ../ocropy/tests/run-unit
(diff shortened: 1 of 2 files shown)


kba:
You don't need the `return 0` statements and if the tests grow more complex, it might be a good idea to factor out the TAP-emitting or switch to `unittests` or similar but for now LGTM.

zuphilip:
> You don't need the return 0 statements

Okay, I guess this is is the standard behavior then...

> ... if the tests grow more complex, it might be a good idea to factor out the TAP-emitting or switch to unittests or similar but for now LGTM.

Yes, this are good ideas to continue to work on unit tests. Let us keep that in mind when we expand the unit tests here.

PR #216, opened 2017-05-08
Delete .hgignore

zuphilip:


stweil:
`.hg` can also be removed from `.gitignore`.

Issue #88, opened 2016-03-04
Not segmenting if image size is less than 600x600


srika91:
The first part of binarization is working correctly for my image.But in the segmentation step if both  or one of the height or width is less than 600x600, ocropus segmentation is not segmenting the binarized image. Is it necessary to have image > = 600x600, because upsampling the image makes the ocr part miserable. And also why is this 600x600 limitation? I am giving a cropped input to ocropus, so resizing is a problem.

kendemu:
I am upscaling 480p video frame to 300dpi and have no problem with the ocr. If you want a high quality upscaling, use [waifu2x](https://github.com/nagadomi/waifu2x).

zuphilip:
In the code you can deactivate this test by changing the [line 385 in toplevel.py](https://github.com/tmbdev/ocropy/blob/master/ocrolib/toplevel.py#L385) to
```py
    return a.ndim==2# and a.shape[0]>=600 and a.shape[1]>=600
```
It might be that this assumption was made also later in the code and therefore some tools or function are not working as expected. If someone has deeper insides or maybe experiences with the image size assumption, that would be helpful.

amitdo:
From `ocropus-gpageseg`:

```
def check_page(image):
    if len(image.shape)==3: return "input image is color image %s"%(image.shape,)
    if mean(image)<median(image): return "image may be inverted"
    h,w = image.shape
    if h<600: return "image not tall enough for a page image %s"%(image.shape,)
    if h>10000: return "image too tall for a page image %s"%(image.shape,)
    if w<600: return "image too narrow for a page image %s"%(image.shape,)
    if w>10000: return "line too wide for a page image %s"%(image.shape,)
    ...
```

tmbdev:
The text recognizer requires 300 DPI, and the page segmenter requires full pages from articles or books. The check is there because the kind of image you give it doesn't conform to these requirements and you need to do something to make it work.

tmbdev:
You could try to upsample your image and then put it into the center of a large blank page so that both requirements are met. Unfortunately, there are multiple parameters involved (resolution and page size), so it requires since intelligence to figure out want to do.

Future recognizers will try to deal with this more intelligently.

amitdo:
Tom,
Does your new project will be written in Python or C++ or maybe in Lua ? Are you going to use a framework like TensorFlow / Theano / Torch ?

tmbdev:
[...]

PR #220, opened 2017-05-14
Change checks for write_page_segmentation

zuphilip:
This seems to solve #88 and allows to run page segmentation also in smaller images (less than 600x600), by using the `-n` option which deactivate the checks. The function `write_page_segmentation` should IMO not depend on the size of the image. Moreover, the check is now equal to the one for the function `read_page_segmentation` (cf. some lines above).

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -279,7 +279,6 @@ def read_page_segmentation(fname):
     segmentation = make_seg_black(segmentation)
     return segmentation
 
-@checks(str,PAGESEG)
 def write_page_segmentation(fname,image):
     """Writes a page segmentation, that is an RGB image whose values
     encode the segmentation of a page."""


ocrolib/common.py
```diff
@@ -279,7 +279,7 @@ def read_page_segmentation(fname):
     segmentation = make_seg_black(segmentation)
     return segmentation
 
-@checks(str,PAGESEG)
+@checks(str,_=PAGESEG)
```
zuphilip: What is the meaning of this change? Help...

kba: checks is here https://github.com/tmbdev/ocropy/blob/ace19d5801c2dcd4d669ad4fac53b495e5077fe8/ocrolib/toplevel.py#L195

It expects positional arguments and keyword arguments that map argument names to types which are checked (I think). It can also check the **return** type agains the keyword argument `_` (https://github.com/tmbdev/ocropy/blob/ace19d5801c2dcd4d669ad4fac53b495e5077fe8/ocrolib/toplevel.py#L218).

So your change makes a check that `write_page_segmentation` returns `PAGESEG` (https://github.com/tmbdev/ocropy/blob/ace19d5801c2dcd4d669ad4fac53b495e5077fe8/ocrolib/toplevel.py#L428). Which it does not, it retuns `None`.

In short, I think you could just omit the `checks` decorator :whale:

zuphilip: I don't follow here. AFAIK the change I make is deactivating the check for PAGESEG which is kind of conform to the other deactivated checks at this point. Thus, I think it should be okay to merge this PR. Or do you suggest to delete this line completely?

kba: I think the check does nothing so I would remove it.

zuphilip: Okay, fine for me. There might be more of this sort of checks...

kba: Yeah, it's not ideal because the original check did make sense (checking that the second argument, `image` is one of `PAGESEG`.

PR #233, opened 2017-07-16
Ignored empty lines in fonts list file

ChillarAnand:


kba:
That is much more idiomatic.

Issue #208, opened 2017-04-28
Detection of black column separators


zuphilip:
Sometimes the detection of black column separators looks very wrong. I have experienced some black lines which are (wrongly) extended until the end of the page. Moreover, it looks that sometimes not the (globally) best candidate(s) are chosen. Are these some systematic errors which can be fixed? Or is the current algorithm just too fragile? However, on some other images the black line detection is reasonable well.

## Example 1: Flugschrift

```
 /etc/ocropy/ocropus-gpageseg flugschrift-landesarchiv.bin.png -b --maxseps 5 --debug -n --minscale 11
```

![_colseps](https://cloud.githubusercontent.com/assets/5199995/25541995/0f2857ec-2c51-11e7-9b23-c5811d608a81.png)

[This is an example](https://cloud.githubusercontent.com/assets/5199995/25542028/2cc2139c-2c51-11e7-92a2-d8fd06dc0fc4.png) were the lines are extended until the end of the page. I don't know why these wrong lines are detected. Maybe some letters contain vertical lines which are extended too much? Moreover, by choosing only the best candidate `--maxseps 1` the leftmost line is choosen and not the longer line in the middle.

## Example 2: Reichsanzeiger

```
ocropus-gpageseg reichsanzeiger/0001.bin.png -b --maxseps 5 --debug --maxlines 500
```
![_colseps](https://cloud.githubusercontent.com/assets/5199995/25541694/ee726606-2c4f-11e7-956a-dfb03148ed6f.png)

[This is an example](https://cloud.githubusercontent.com/assets/5199995/25541777/4917116a-2c50-11e7-9688-a25121aa91d2.png) were the lines are extended until the end of the page. Actually the outer two separators would be only a few lines hight and therefore neglected by better candidates. Moreover, the missing separator on between the second and third column was deleted at some point.

## Example 3: Twocolums

[This is an example](https://cloud.githubusercontent.com/assets/5199995/25541901/babd4df2-2c50-11e7-9bac-f584fdb49fed.png) were the line detection works reasonable well:
```
ocropus-gpageseg twocolumns.bin.png --debug -n --maxseps 3 -b
```

![_colseps](https://cloud.githubusercontent.com/assets/5199995/25541611/8c7c6488-2c4f-11e7-9a6a-e09ffc70d027.png)

Issue #89, opened 2016-03-08
Black column separator 


1a1a11a:
I have pages with three columns separated by black column separator, but sometimes during the process, the black lines are recognized as character, does any one have idea how to solve that? Besides, sometimes two lines are also recognized as one line, is there any solution on that?

jze:
Did you activate the detection of black column separators using the -b option? You also have to specify the number of black separators with the --maxseps option. The default is 2. However, if no continuous black line is detected, the specified maximum number of  black separators must be higher than the actual number.

1a1a11a:
Thank you for your reply! I used both -b and --maxseps options, but I don't quite understand what you mean by "the specified maximum number of black separators must be higher than the actual number". I have three black columns each separated by a black vertical line, but the vertical line is occasionally partial disconnected (historical documents, ~100yeas, so sometimes and black line is not that clear). I can understand the program might connect two columns if the separator is not clear, but sometimes, it will just treat the black separator as characters 'l, i, 1' and I don't know why. Can you help? @jze Thank you!

1a1a11a:
I am not sure whether I stated it clearly, if not, just let me know, maybe I can find some examples.

jze:
Have you tried the debug mode -d ? With debug enable ocropus-gpageseg writes an image _colseps.png. In this image the detected column separators are highlighed. I have not found the parameter that controls how disconnected vertical black lines might be to be treated as a continuous line. In an experiment a hole of 5 pixels cut the line in two. A 4 pixel hole was ignored. (scale was 19.079)

1a1a11a:
I haven't try debug mode, let me try it, sounds like a good suggestion. In your experiment, do you know how to control the number of pixels (sounds like a threshold) for cutting?

sepastian:
Hello, I am facing a similar problem with historic documents.

The black lines separating columns are sometimes disconnected. Is there a way of specifying a threshold saying, in essence, ignore holes in black lines smaller than X pixels?

I'm attaching two imgaes, showing the binarized text and the columns detected in debug mode with `./ocropus-gpageseg --maxcolseps 0 -b --maxseps 10 --sepwiden 50 -d -n 0001.bin.png`.

![0001 bin](https://cloud.githubusercontent.com/assets/14915/22975770/13d141ca-f388-11e6-8e6e-dfb9eaa089c1.png)

[...]

Issue #38, opened 2015-05-16
Avoid text recognition in image areas of the page


zuphilip:
In several tests ocropus/ocropy tries to recognize some text within an image area in my pages. How can this be avoided? Here is an example of such a page: [normal-beispiel-mit-bild](https://cloud.githubusercontent.com/assets/5199995/7665570/53263dda-fbbe-11e4-8162-e9d7c7ec56b8.png)

I run the same sequence of commands as in `run-test` and ocropus/ocropy recognizes two columns one with the actual text and the other with some nonsensical symbols from the picture.

tmbdev:
Generally, that requires good text/image segmentation, which isn't part of ocropy yet.

The plan is to provide a 2D LSTM implementation to learn text/image segmentation. For that that, I've been working on the CLSTM implementation.

zuphilip:
Okay, I see. Can you say something about an ETA for such a feature? Days, weeks, months?

tmbdev:
It's probably going to be months; it's a pretty hard problem to solve in general, and there is only limited training data.

Here is a paper describing an older approach: http://www.csse.uwa.edu.au/~shafait/papers/Bukhari-Text-Image-Segmentation-DAS10.pdf

kungfujam:
I'm doing my masters thesis on this topic at the moment and would like to get involved in implementing segmentation useful for ocropy. Are you looking for people to help?

ankitag9:
@tmbdev @danvk I have been using ocropy for some time now and have made a few changes as well.. I am planning to work on ML based image segmentation next.. do you guys have some alpha version code of the same on which i can build?

ChillarAnand:
Any updates on this issue?

zuphilip:
Old C code is here https://github.com/michaelyin/ocropus-git/blob/ba6930627f3f81392a05a02bfcf7dac2595c35cf/ocr-layout/ocr-text-image-seg.cc and some Python code (~2012) https://github.com/zuphilip/ocropus-from-searchcode/blob/master/ocropy/ocropus-tiseg resp. https://github.com/zuphilip/ocropus-from-searchcode/blob/master/ocropy/OLD/ocropy/oldsimpleti.py

amitdo:
Both Tesseract and ocropus 0.4 use the Leptonica library to do text/image segmentation.

PR #236, opened 2017-07-21
ocropus-gpageseg: Enable usage of masks to specify column separators/ ignore areas of scan

lehzwo:
### Intention

As stated in #171 @jze had the idea of applying a mask to an image if the column separator detection did not work as expected.
This behavior could be interesting for #208 as well as #89. 

### Current behavior of the mask

A white on black mask can be applied to an image file. The white areas of the mask are interpreted as column separators and are merged with the column separators that have been identified by OCRopus itself.
The merged separators are then treated as before. If no mask is found for an image, ocropus-gpagesegs behavior is not changed.

### Naming convention

The file which specifies the mask is saved as the basename of the file it shall be applied on, extended by ".mask.png".

e.g. the basename of the file is _".../book/0003"_
in this case the mask is stored as _".../book/0003.mask.png"_

### Further usages

The current behavior of column separators identified by OCRopus is, that they delete the pixels beneath them. The column separators specified by the mask currently behave the same way. Due to this the mask can also be used to ignore areas of pixels of a scan. **This might screw up the segmentation, though!** 

This way images as in the example of #38 or other areas could be ignored.

### Potential improvements/ changes

The usage of the mask to ignore areas might lead to segmentation problems if not used with care. Due to this it might be desirable to be able to apply different kind of masks like. A mask that is extended by ".mask.sep.png" for column separators and another mask ".mask.ia.png" that ignores areas of the scan.
This would then lead to multiple mask files per image. Another way to achieve this behavior might be to work with different colors. In this case only one mask file is used. This would work for the actions "ignore area" and "apply separator" but is not as extendible since each pixel can only have one color.

Another potential improvement would be to introduce a new parameter to ocropus-gpageseg where a "master-mask" _(/master mask**s**, if multiple masks are allowed)_ can be passed that is applied to every image, processed by ocropus-gpageseg. This way unwanted recurring patterns can be easily ignored on each scan without renaming/ copying an existing mask.

### Examples 

Mask for first image of #208 as well as lineseeds with and without mask. Mask is used to identify correct separator and prevent identification of wrong separators in the text of the scan.
Params:
_ocropus-nlbin -b 0.2
with mask:
    ocropus-gpageseg --debug -n --minscale 11_
without mask: 
    ocropus-gpageseg -b --maxseps 5 --debug -n --minscale 11

Disabled colum separator identification by OCRopus when the mask is used since we only want to apply the separators specified by the mask in this case.

0001 mask
lineseeds_applied_mask
lineseeds_without_mask

Mask for image of #89 as well as lineseeds with and without mask. Mask is used to prevent "holes" in the separators _(would have been enough to mark holes in the mask instead of complete separators since the specified areas are merged with the separators found by OCRopus)_.
Params: 
_ocropus-nlbin -b 0.2
ocropus-gpageseg --maxcolseps 0 -b --maxseps 10 --sepwiden 50 -d -n_

0001 mask
lineseeds_applied_mask
lineseeds_without_mask

Mask for image of #38 as well as lineseeds with and without mask. Mask is used to ignore image on scan.
Params:
_ocropus-nlbin -b 0.2
ocropus-gpageseg --maxseps 1000 --maxcolseps 1000 -n -d --maxlines 10000_

0001 mask
lineseeds_applied_mask
lineseeds_without_mask


kba:
This is both simple and useful, esp. for images with static layout. Is backwards-compatible too. We should add tests/examples and document it.

zuphilip:
Yes, this looks really interesting! I haven't had the time to look at it. Really sorry for that :-( I will do and feel free to ping me again...

zuphilip:
**tldr**: Thank you very much @lehzwo for this nice PR! ✨ This looks very useful and is just a small addition to the code as @kba already pointed out. We are IMO ready to merge this now. @kba Can you do the final review and merge it then?

---

I updated the PR and add a test case. However, I think that the smoke test here may not detect many errors. Currently, the output of the test results looks like:
```
INFO:
INFO:  ########## ./ocropus-gpageseg temp/table.bin.png --debug -n --minscale
INFO:
INFO:  temp/table.bin.png
INFO:  scale 7.416198
INFO:  computing segmentation
INFO:  computing column separators
INFO:  considering at most 0 whitespace column separators
INFO:  debug _1thresh.png
INFO:  debug _2grad.png
INFO:  debug _3seps.png
INFO:  debug _4seps.png
INFO:  debug _colwsseps.png
INFO:  debug _masked_seps.png
INFO:  computing lines
INFO:  debug _cleaned.png
INFO:  debug _lineseeds.png
INFO:  debug _seeds.png
INFO:  propagating labels
INFO:  spreading labels
ERROR:  temp/table.bin.png: too many lines 613
```
I didn't adjust the maximum number of lines, because the ordering of the lines takes quite some time. Thus, it results with this ERROR, but still a valid test case for us.

> The current behavior of column separators identified by OCRopus is, that they delete the pixels beneath them. The column separators specified by the mask currently behave the same way. Due to this the mask can also be used to ignore areas of pixels of a scan. This might screw up the segmentation, though!

Yes, this another nice application of such masks. I cannot think of any negative effect for the segmentation. The "smearing" step will not go into the white areas of the mask, but everywhere else where it would go in normal run as well. This looks like exactly the behavior we want. @lehzwo Did you have anything specific in mind here? I rather would go further from here: with these "masks" it is also possible to run the recognition only in a limited area by just coloring everything else white. Thus, one can for example recognize only one column and ignoring the remaining parts of the page.

The only concern I have at the moment is, that this feature is now very hidden. We certainly should document it well, e.g. in the wiki https://github.com/tmbdev/ocropy/wiki/Page-Segmentation . But maybe it also makes sens to add an additional parameter for it. With an additional parameter we can then change the default name as you suggest, but more important IMO is that the feature appears in `--help`. But maybe we can do this later in another PR...

Issue #197, opened 2017-03-27
'UnicodeEncodeError' when redirecting status outputs


wrznr:
## Expected Behavior
I'd like to be able to redirect the print statements excuted during e.g. training to a file. This is also necessary for using the <nohup> tool for running ocropus-* as a background process.

## Current Behavior
When redirecting the status output of ocropy's training or linegen scripts to a file (either directly or while using <nohup>), non-ASCII characters cause a UnicodeEncodeError exception ultimately stopping the script.

## Possible Solution
print statements should use explicit encoding to utf-8 to make sure they can be safely printed to various destinations.

## Steps to Reproduce (for bugs)
 
![010000 bin](https://cloud.githubusercontent.com/assets/26707219/24359292/3588ec3c-1304-11e7-8e68-0a53bd091fe4.png)
[010000.gt.txt](https://github.com/tmbdev/ocropy/files/872593/010000.gt.txt)

1. Download image and gt text.
2. ocropus-rtrain -o test --codec=bug/010000.gt.txt bug/*.png 1> status.txt

## Your Environment
* Python version: Python 2.7.3

* Git revision of ocropy: 
commit fc525cb2fc07d7d8e29bbcf9fa3b7989d59df8cc
Merge: 4efbddc ae763b0
Author: Philipp Zumstein <zuphilip@users.noreply.github.com>
Date:   Thu Mar 23 08:32:46 2017 +0100

* Operating System and version:
Debian
7.9

kba:
To get this running until fixed properly, try setting the `PYTHONIOENCODING` to `utf-8`, e.g.

```
export PYTHONIOENCODING=utf-8
ocropus-rtrain -o test --codec=bug/010000.gt.txt bug/*.png 1> status.txt
```

That should prevent python from wrongly decoding utf-8 codepoints to latin1 bytes.

wrznr:
Many thanks! Nice workaround. Should we still call this an issue (i.e. "until properly fixed")?

amitdo:
The Hebrew text in your image is reversed (mirrored).

ocropy doesn't deal well with RTL scripts out-of-the-box, although you probably can make it work with some tricks.

wrznr:
Thanks amitdo. bidi.algorithm turns out to be the perfect choice for dealing with that problem. However, what I was asking: Is kba's workaround a proper fix for the issue filed here?

kba:
[...]

Issue #10, opened 2014-12-03
ocropus-rpred crashes when printing unicode text without locale


mittagessen:
When running ocropus-rpred without a locale (e.g. using subprocess.Popen) and recognizing non-ASCII character it tries to print them directly onto the command line causing an UnicodeEncodeError and subsequent skipping of the line. The offending line is 197 in ocropus-rpred:

```
   print fname,":",pred
```

As similar calls appear to occur in a wide range of ocropus utilities it would be sensible to wrap them all in the correct encode() statements.

PR #237, opened 2017-07-28
Explicitly encode output as UTF-8

nickjwhite:
I think I've covered all the cases we need to, but it's possible there could be more.

This should address issue #197 and #10.


kba:
These will only change the debug output. Python3 should already do this but won't break with it. LGTM. @zuphilip ?

QuLogic:
`str` + `bytes` does not work; this is broken on Python 3.

nickjwhite:
Thanks @QuLogic, I don't write a lot of Python, so hadn't spotted the mixing of str + bytes.

I just updated the pull request, with a version that ensures nothing but str is in the print() arguments. And to force UTF-8, I switched to using codecs.getwriter('utf8') for stdout and stderr for all cases where they could output it. I also improved the commit message, so it's clearer what the problem is and why it's worth fixing.

nickjwhite:
Just as an additional justification / use-case for this being necessary, I'm currently using Ocropus on a system (that I don't manage) which uses slurm, and no matter what I try I cannot persuade the slurm job manager on this system to not strip locale information. And again, to reiterate, UTF-8 is always the correct thing to output here, regardless of locale, and no fatal errors should be emitted just because of a weird locale.

QuLogic:
Does setting ['PYTHONIOENCODING`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING) not help?

kba:
As laid out in #197 there are a few options to achieve this. Revisiting the UTF8/debug output issues, I'm not keen on explicitly adding boilerplate to every file with print statements. Changing stdout/stderr encoding in a single place (like common.py) probably achieves the same goal but is error-prone.

I would prefer to replace print/print_info/print_error with a dedicated logging mechanism, such as python's bundled `logging` or something smaller in the codebase. That would make it easier to override behavior (such as output stream, encoding, logging level).

Issue #251, opened 2017-10-16
Regression in ocropus-gpageseg


jze:
The changes between rev 2e83dcb and 8d378ab caused a regression in ocropus-gpageseg. The older version produces a line like this:
![good nrm](https://user-images.githubusercontent.com/765178/31597611-a295fd1c-b249-11e7-872c-1c5d7bb86ec7.png)

With the new version a line includes some pixels from the next line:
![bad nrm](https://user-images.githubusercontent.com/765178/31597606-9d000b04-b249-11e7-8df9-8d1cdf81d662.png)

stweil:
Could you try to narrow it down to the commit which causes the regression by using `git bisect`?

jze:
2e83dcb and 8d378ab are successive commits for `ocropus-gpageseg`

stweil:
Yes, but 8d378ab0e9dd832638c0205149fd2cf6165215b7 merged several commits. How can I reproduce the problem myself?

zuphilip:
Fixed in #252

PR #252, opened 2017-10-16
Fix regression in ocropus-gpageseg #251

jze:
I looks as if pylab.find and np.where work slightly different.

### DIFF
--- a/ocropus-gpageseg
+++ b/ocropus-gpageseg
@@ -13,6 +13,8 @@
 
 from __future__ import print_function
 
+from pylab import find
+
 import argparse
 import glob
 import os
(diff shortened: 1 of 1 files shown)


amitdo:
https://github.com/tmbdev/ocropy/commit/a154e6360cd8#diff-0e349d933e2d87621d932d2ae80cf70aL272

QuLogic:
Please do not start using the `pylab` interface again.

kba:
Thanks for reporting and sorry for the regression.

[IIUC](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/mlab.py#L1547-L1550), `pylab.find` does this:

```python
def find(condition):
    "Return the indices where ravel(condition) is true"
    res, = np.nonzero(np.ravel(condition))
    return res
```

Does `s,np.find/np.nonzero(np.ravel,` fix the issue?

jze:
Yes, it also works without pylab.

kba:
LGTM, thank you.

@zuphilip In https://github.com/tmbdev/ocropy/commit/a154e6360cd8e15c8c7cd5c2ce225c0ebed9d424 this was the only place where pylab.find was replaced by np.where, right?

@jze If you provide the source images and CLI invocation, we could add a test to prevent this kind of error in the future

stweil:
> In a154e63 this was the only place where pylab.find was replaced by np.where, right?

Yes.

zuphilip:
Merged. Thank you very much @jze for the work and everybody here for the discussion!

PR #260, opened 2017-11-16
Changed method 'read_image_gray()' and 'read_image_binary()'. The old…

jingchaoluan:
Goal: Extend the ability to process image only from disk to from both disk and memory.

Updates: changed methods 'read_image_gray()' and 'read_image_binary' in module 'ocrlib/common.py'. 

Description: The method can only read an image from disk i.e. from a filename (string). The new method extends the ability to read an image from both disk and memory. Read image from memory means that it can also accept file object or PIL.Image.Image object as the parameter of this method.

My application context: Now I developed OCRopus related services and used the library 'ocrlib', but it can only read an image from disk (the parameter type must be 'string'). While for my service, I used the Django framework, which receives the image uploaded by users as an 'InMemoryUploadFile' type which is a file type object. Thus, the 'ocrlib' can not take it as a parameter of the read_image_ methods and thus cannot process. So I changed the methods and now it can process image from disk (string type) and memory (file object or PIL.Image.Image)

### DIFF
--- a/ocrolib/common.py
+++ b/ocrolib/common.py
@@ -152,7 +152,7 @@ def isintarray(a):
 def isintegerarray(a):
     return a.dtype in [dtype('int32'),dtype('int64'),dtype('uint32'),dtype('uint64')]
 
-@checks(str,pageno=int,_=GRAYSCALE)
+@checks({str,object},pageno=int,_=GRAYSCALE)
 def read_image_gray(fname,pageno=0):
     """Read an image and returns it as a floating point array.
     The optional page number allows images from files containing multiple
@@ -160,7 +160,10 @@ def read_image_gray(fname,pageno=0):
     the range 0...1 (unsigned) or -1...1 (signed)."""
     if type(fname)==tuple: fname,pageno = fname
     assert pageno==0
-    pil = PIL.Image.open(fname)
+    if type(fname) == PIL.Image.Image:
+        pil = fname
+    else:
+        pil = PIL.Image.open(fname)
     a = pil2array(pil)
     if a.dtype==dtype('uint8'):
         a = a/255.0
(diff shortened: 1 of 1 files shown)


kba:
I think this is a useful extension, thanks.

Can you create a short python script that tests this behavior, i.e. passing a `PIL.Image` and a file name to the functions and having the same result? This can be really short and should use the assets in the test directory. Then we can this as a unit test.

jingchaoluan:
Thank you. @kba 

I write a simple script to only call methods "read_image_gray()" and "read_image_binary()" with a parameter 'file name' and 'PIL.Image.Image instance' respectively. And print their return value (array type) to compare. 

I didn't put these two methods in image processing to compare their output, in which I need to change the script 'ocropus-nlbin' and 'ocropus-gpageseg' to call these two methods with PIL.Image.Image instance respectively. Certainly, I also test them in local, if you need I can upload the script and updated scripts 'ocropus-nlbin' and 'ocropus-gpageseg' which only changed two code lines to call these two methods with PIL.Image.Image instance.

Besides, I upload the script in TXT file tyle for the limitation of uploaded file type.

[ocrlib_disk_mem.txt](https://github.com/tmbdev/ocropy/files/1563904/ocrlib_disk_mem.txt)

jingchaoluan:
Hi @kba 

Sorry for replying the unit test so late. The attached file is the unit test code for testing functions read_image_gray() and read_image_binary(). 

Because these two functions accept a raw image and a binarized image respectively, I used two images (tests/testpage.png and tests/010030.bin.png) and their related PIL.Image objects to test. 

You can run this test code by uncomprocessing it under folder 'ocropy', and certainly with the updated read_image_gray() and read_image_binary().

[test_read_image.py.zip](https://github.com/tmbdev/ocropy/files/1696844/test_read_image.py.zip)

PR #264, opened 2017-11-27
Have pageseg output decimal rather than hex numbering

nickjwhite:
This is useful as it means that normal alphabetical ordering of files works as expected. This allows one to do things like `cat *txt >wholepage.txt` to get a complete transcription of a page.


zuphilip:
See #184.

nickjwhite:
> See #184.

My apologies, I hadn't seen that pull request, I'll cancel this one, yours does a better job (not least because it handles pseg / ocropus-hocr usage properly too).

PR #265, opened 2017-11-27
Replace native code with regular functions

amitdo:
Use numpy.einsum() for sumouter and sumprod functions.

### DIFF
--- a/ocrolib/lstm.py
+++ b/ocrolib/lstm.py
@@ -35,7 +35,7 @@ from pylab import (clf, cm, figure, ginput, imshow, newaxis, rand, subplot,
 from collections import defaultdict
 from ocrolib.exceptions import RecognitionError
 from ocrolib.edist import levenshtein
-import nutils
+import utils
 import unicodedata
 from scipy.ndimage import measurements,filters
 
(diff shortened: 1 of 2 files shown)


nickjwhite:
Looks good. I wonder whether the native code was originally used because these functions were slow through python? I suspect not, and this is a worthwhile change.

I actually just found a bug with the native compiler today, as '-fopenmp' wasn't a defined flag on the OSX compiler I was using. But this fixes it in a nicer way :)

If I'm not mistaken, this is the only place the nutils are used, so why not remove `ocrolib/nutils.py` and `ocrolib/native.py` too?

amitdo:
Hi Nick!

> I wonder whether the native code was originally used because these functions were slow through python?

I don't know... :-) 

OpenMP
https://github.com/tmbdev/ocropy/blob/master/README_OSX.md

Actually, there is no actual OpenMP 'code' in nutils, so the '-fopenmp' flag is useless.

>If I'm not mistaken, this is the only place the nutils are used, so why not remove ocrolib/nutils.py and ocrolib/native.py too?

Originally, I removed that code, but for this PR I decided to keep it. Later, it can be removed or moved to 'OLD'.

amitdo:
In my mini benchmarks the new functions are a few times faster than the native ones in nutils.

I ran ocropus-rtrain with '-N 1000'  on uw3-500.tgz and it looks like there is a ~20% improvement in speed for training. I think it's a good idea that someone else will verify this in his machine.

amitdo:
In lstm.py there are another sumouter and sumprod function.

The sumprod in lstm.py is not used in the code.

The sumouter is used in the softmax layer. For some reson it operates on lists instead on numpy ndarrays.

kba:
Thank you @amitdo. If you want to start a PR with the rest of your [dropnative](https://github.com/amitdo/ocropy/tree/dropnative) branch changes, feel free.

Issue #269, opened 2017-12-08
Remove remnants of native/code generator functionality 


kba:
With #265 closed, we can drop `nutils.py` and `native.py`.

Also `mkpython` and `pyconstruct` are obsolete.

README for OSX not required anymore.

Installation instructions should reflect no need for gcc anymore.

kba:
Merged in #274

PR #270, opened 2017-12-09
Bump version to 1.3.2

zuphilip:
New tags/releases were created now retroperspectively, but we should try to continue that better just in time. With the new commit which gets rid of the native code, we can plan a new version 1.3.2.

This is on hold until #269 is solved. Any more testing of the current state would also be appreciated.

(After merging this commit can then be tagged with `v1.3.2`.)

### DIFF
--- a/ocrolib/hocr.py
+++ b/ocrolib/hocr.py
@@ -14,7 +14,7 @@ header_template = """\
 <head>
 <title>OCR Results</title>
 <meta http-equiv="content-type" content="text/html; charset=utf-8" />
-<meta name="ocr-system" content="ocropy-1.0" />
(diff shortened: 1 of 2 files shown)


kba:
#269 is merged. I drafted a release. I'd include #276 because it cannot hurt and #275 because it's simple and useful.

kba:
I released [1.3.2](https://github.com/tmbdev/ocropy/releases/tag/v1.3.2) but missed this PR, newest release is [1.3.3](https://github.com/tmbdev/ocropy/releases/tag/v1.3.3) with the version numbers set for hocr output and in setup.py.

Issue #210, opened 2017-05-02
ocropus-gpageseg: Defective line splitting


wrznr:
## Expected Behavior
Simple running text should be consistently split into lines.

## Current Behavior
Currently working on data from the [Grenzboten project](https://www.suub.uni-bremen.de/ueber-uns/projekte/grenzboten/) together with @uvius. For some images, line splitting does not work. It is not clear why because very similar images are split correctly.

## Steps to Reproduce (for bugs)
1. Download test files: 

![179411_01 nrm](https://cloud.githubusercontent.com/assets/26707219/25614998/063f51b2-2f35-11e7-91ec-c11b55e4e5d5.png)
![179411_01 bin](https://cloud.githubusercontent.com/assets/26707219/25614999/063fa2c0-2f35-11e7-9c83-0cb354a42522.png)

2. Run `ocropus-gpageseg` on testfile(s).
3. Inspect results.

Test files have been created with `ocropus-nlbin`. Tested various command line parameter settings without success.

## Your Environment
* Python version: Python 2.7.3
* Git revision of ocropy:  commit 49c7f9ec943bda2c249b9027776d0e3de477b5ec
* Operating System and version: Linux lal 3.2.0-4-amd64 #1 SMP Debian 3.2.73-2+deb7u2 x86_64 GNU/Linux

zuphilip:
Okay, I looked at the debug output with ` --debug` and it seems that the detected scale is too small (approximately half of the correct size):

![scale-default_lineseeds](https://cloud.githubusercontent.com/assets/5199995/25630565/e9cd66a2-2f6d-11e7-89c8-1b600c7348b9.png)

The disconnected (red) components are then creating the different lines.

If you increase that value by hand by setting the `--scale` parameter:
```
ocropus-gpageseg grenzboten.bin.png -n --debug --scale 30
```
then the output looks good:

![scale-30_lineseeds](https://cloud.githubusercontent.com/assets/5199995/25630660/2e3a6ac4-2f6e-11e7-95da-b37d53899262.png)

(Don't forget to remove all old images from the directory containing the lines.)

wrznr:
Alright, thanks. That fixes the issue for the specific image (and many others). But if I set this parameter globally for the whole (pre-segmented) book, new problems arise with smaller (e.g. on-line images). Is there a known bug in the scale detection?

zuphilip:
> new problems arise with smaller (e.g. on-line images)

[...]

PR #273, opened 2017-12-09
Delete unused lru.py

zuphilip:


kba:
This one we could actually delete, since it's just a copy of the snippet from http://code.activestate.com/recipes/498245-lru-and-lfu-cache-decorators/.

zuphilip:
First, I wanted to delete completely, but then I saw that there is code in `OLD` which is relying on this `lru.py`: https://github.com/tmbdev/ocropy/search?utf8=%E2%9C%93&q=lru&type= and thus I moved it also to OLD. As you see, these are not strong reasons against deleting completely. To be honest, I have never used any file or code from `OLD` so far...

amitdo:
About the OLD dir. Maybe it should be removed.

Git keeps the history, you know...

amitdo:
'ocropus-sauvola' might be useful. See #210.

Consider moving it back to the root dir. It could be used as an alternative to n
'ocropus-nlbin'.

amitdo:
>but while you can find/retrieve old files in Git if you know how and what to look for, discovering useful deleted code is tedious (e.g. when searching in Github or Google).

You can solve this issue by creating a new 'ocropy-unused' repo and move most of the stuff in 'OLD' dir to this repo.

zuphilip:
Okay, the file is deleted now. @kba ready to merge?

PR #274, opened 2017-12-11
Native cleanup

kba:
This realizes #269, dropping support for on-the-fly-compiled C code, some dead code in common.py and references in the docs how to get compilation to work with clang on OSX.

It does not touch sumouter/sumprod beyond what #265 did. @amitdo/@nickjwhite: If you want to refactor lstm.py further as discussed in https://github.com/tmbdev/ocropy/pull/265#issuecomment-347275497 it would be appreciated!


amitdo:
lstm.py: 

```
def sumprod(us,vs,lo=-1.0,hi=1.0,out=None):
    """Sum the element-wise products of the `us` and `vs`.
    Values are clipped into the range `[lo,hi]`.
    This is mainly used for computing weight updates
    in logistic regression layers."""
    assert len(us[0])==len(vs[0])
    result = out or zeros(len(us[0]))
    for u,v in zip(us,vs):
        result += clip(u,lo,hi)*v
return result
```

It's not being used anywhere, so you can safely remove it.

ocrolib/__init__.py
```diff
@@ -1,7 +1,7 @@
 __all__ = [
```
amitdo: This list is outdated. You can do more cleanup here.

kba: Is that just used for `from ocrolib.xyz import *`? That idiom is not used in ocropy's code base itself. I would not want to break any tools relying on that but I would not add to the list, merely remove old modules.

amitdo:
>remove sumprod from common.py (in favor of utils.sumprod)

It's from tstm.py, not  common.py
Just to make it clear, the local sumprod() was **never** used, unlike the one from nutils (and now utils).

:-)

ocrolib/__init__.py
amitdo: What you did is OK, it's what I meant.

kba:
> It's from lstm.py, not common.py

True, I changed that. I have another branch here with just deletions from common.py, that was a typo :blush: 

Thanks for reviewing!

Issue #93, opened 2016-05-05
Error in ocropus-rtrain


anupamaray:
I am trying to use ocropus-rtrain: ocropus-rtrain -o model try/*.bin.png
At the begining itself I am getting this error (I am using Ubuntu ans all images with gt.txt files are in try folder)

`# oops, got FloatingPointError divide by zero encountered in double_scalars
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-rtrain", line 286, in <module>
    pcs = network.trainSequence(line,cs,update=do_update,key=fname)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/lstm.py", line 902, in trainSequence
    self.error_log.append(self.error**.5/len(cs))
FloatingPointError: divide by zero encountered in double_scalars
Traceback (most recent call last):
  File "/usr/local/bin/ocropus-rtrain", line 290, in <module>
    network = load_lstm(last_save)
  File "/usr/local/bin/ocropus-rtrain", line 176, in load_lstm
    network = ocrolib.load_object(last_save)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/common.py", line 503, in load_object
    fname = ocropus_find_file(fname)
  File "/usr/local/lib/python2.7/dist-packages/ocrolib/common.py", line 682, in ocropus_find_file
    if os.path.exists(fname):
  File "/usr/lib/python2.7/genericpath.py", line 18, in exists
    os.stat(path)
TypeError: coercing to Unicode: need string or buffer, NoneType found`

 Can someone help with this error?
@kba please let me know if some mode detail would help.  I have sample data, command lines and model files if they would helpful.

kba:
The problem is in `ocropus-rtrain`, line 248

```python
     cs = array(codec.encode(transcript),'i')
```

If cs has zero length, this will lead to a division by zero in `ocrolib/lstm.py`, line 902:

```python
        self.error_log.append(self.error**.5/len(cs))
```

The problem is `transcript`, which is loaded in line `ocropus-rtrain`, line 263:

```python
    base,_ = ocrolib.allsplitext(fname)
    try:
        line = ocrolib.read_image_gray(fname)
        transcript = ocrolib.read_text(base+".gt.txt")   ### <- HERE
    except IOError as e:
        print "ERROR",e
        continue
```

Are you sure your transcripts (`*.gt.txt`) are there and not empty?

anupamaray:
[...]

PR #275, opened 2017-12-12
Added option to load training data through file

Beckenb:
Together with @uvius I added the option to pass the training images through a file, avoiding the "Argument list too long" error when passing too many files.

This allows the user to pass as many files as he wants to ocropus-rtrain, avoiding the `argument list too long` error of the shell, as in Issue #93.


zuphilip:
Thank you @Beckenb and @kba !

Issue #186, opened 2017-02-21
Bringing back ocropus-lattices


lunactic:
Is there a plan to bring back some form of implementation of the ocropus-lattices tool?

It would be great to have the possiblity to extract the recognition-lattices to combine them with an additional language model to possibly improve recognition results.

zuphilip:
I am not aware of that anyone is working on this at the moment. But, I think, Tom outlined how this could be work again in https://github.com/tmbdev/ocropy/pull/25#issuecomment-72075445 . Is this the same issue?

lunactic:
The problems you face when you only search in Issues and not in PRs :-D

But yes, basically it is the same problem. 

If you want to use a language model in cooperation with the recognition output you need to have the recognition-lattices to combine the possibilities of the language model with the possibilities of the recognizer.

It seems this would require quite some work though and I don't know how high up on your todo list something like this would be,

zuphilip:
The old code is still there https://github.com/tmbdev/ocropy/blob/master/OLD/ocropus-lattices , but I have no idea what would be needed to adapt or if one should write it new from scratch. I guess, it would be nice to have such a lattice recognizer in combination with language modeling, but this is, I am afraid, not on my todo list. I don't understand enough about these neural networks to even dare to start here something ;-) Maybe, this could be suitable for some student work?

PR #279, opened 2017-12-15
Implemented computation of probability matrix

BingqingQu:
This is intended to be an extension of the --probabilities.
Instead of just printing the probabilities for the recognised characters, --probmat will compute the complete probability matrix.

At each "timestep" the probability for each character is computed.
This can/could be used as input to a language model for example where one would have access to the probabilities of other characters as well.

### DIFF
--- a/ocropus-rpred
+++ b/ocropus-rpred
@@ -44,6 +44,8 @@ parser.add_argument('--alocs',action="store_true",
                     help="output aligned LSTM locations for characters")
 parser.add_argument('--probabilities',action="store_true",
                     help="output probabilities for each letter")
+parser.add_argument("--probmat",action="store_true",
+                    help="output and save probability matrix after LSTM Network")
 
 # error measures
 parser.add_argument("-r","--estrate",action="store_true",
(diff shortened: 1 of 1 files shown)


zuphilip:
Is your code complete? It looks that the variables `out` and `timestamp` are not used further...

Can you give more information about the output format? I see that the files have always 156 lines with several probalities, but none of these values seem to be equal the ones which are outputed with `--probabilities`.

amitdo:
https://github.com/tmbdev/ocropy/wiki/OCRopus-File-Formats#lattice-files
This format was used in ocropy 0.6.

zuphilip:
@amitdo The outputed files look differently. Here is an example:

[010001.pm.txt](https://github.com/tmbdev/ocropy/files/1564978/010001.pm.txt)
[010001.prob.txt](https://github.com/tmbdev/ocropy/files/1564979/010001.prob.txt)

amitdo:
His patch just outputs the raw result of the prediction.

What you see with the current text/prob. options is the the 'best' path that translate_back() found for you.

The format in my link is more human readable and contains character positions. 


Sorr

amitdo:
Related: #25

amitdo:
The number of lines (156) is the size of the codec (chars) in the model you use.

zuphilip:
Okay, I don't think that this matrix is then enough interesting for an option to `ocropus-rpred`. One can use `ocrolib` as a library for such computations. More advanced lattice/alternative calculations could be interesting as outlined in #186.

zuphilip:
There is also the `--save` and `--show` option for a visual debug info about these matrix.

PR #280, opened 2017-12-16
Add Dockerfile

zuphilip:
I adapted the file from https://github.com/kba/docker-ocropy and it works fine for me under windows especially for developing and testing.


zuphilip:
Let us add `coverage` (and therefore `pip`) as well...

PR #282, opened 2017-12-19
Standardize imports II, #176

zuphilip:
CC @Hyper-Node


ocrolib/edist.py
```diff
@@ -23,10 +24,10 @@ def xlevenshtein(a,b,context=1):
     n, m = len(a), len(b)
     assert m>0 # xlevenshtein should only be called with non-empty b string (ground truth)
     if a == b: return 0,[] # speed up for the easy case
-    sources = empty((m+1,n+1),object)
+    sources = np.empty((m+1,n+1),object)
     sources[:,:] = None
-    dists = 99999*ones((m+1,n+1))
-    dists[0,:] = arange(n+1)
+    dists = 99999*np.ones((m+1,n+1))
```
QuLogic: `np.full((m + 1, n + 1), 99999)`

amitdo:
In lstm.py this form is used:

>from numpy import (...)

This patch uses the np namespace.

Why the difference?

QuLogic:
This one is a de facto standard; I dunno why `lstm` was changed the other way.

zuphilip:
Agree on what @QuLogic says, see also #206.

ocrolib/edist.py
zuphilip: Thank you!

amitdo:
I don't have a problem with using the np namespace, but just wanted to point to the inconsistency.

zuphilip:
Please have a look at the new version of this PR. This should be ready to review now.

(There are more files where the imports could be standardized, but maybe we can first make better tests for them. This is more for some future PRs.)

kba:
`numpy` should be `np` in https://github.com/tmbdev/ocropy/pull/282/commits/23ebc40b0d1ba0dafcea29069b1a231d8df742eb @zuphilip

PR #288, opened 2017-12-27
Use coverage erase and add htmlcov to .gitignore

zuphilip:


run-coverage
```diff
@@ -10,7 +10,7 @@ except ImportError:
     print('\nYou need to install the coverage module first, e.g. \n$ pip2 install coverage\n')
     exit(1)"
 
-rm -rf .coverage*
+$COVERAGE erase
 PATH=$PWD:$PATH $COVERAGE run -p --include=$PWD/ocropus-*,$PWD/**/* ./tests/run-unit
```
zuphilip: Currently, the problem is that the `import ocrolib` in `./tests/run-unit` will use the files in `/usr/local/lib/python2.7/dist-packages/ocrolib/` and not the local one. I guess this is different for the `ocropus-*` scripts, because they are one directory level up. @kba Any ideas how to fix this? Do we have to adjust the PATH variable here?

run-coverage
```diff
@@ -10,7 +10,7 @@ except ImportError:
...
 PATH=$PWD:$PATH RUNNER="$COVERAGE run -p --include=$PWD/ocropus-*,$PWD/**/*" ./run-test-ci
```
zuphilip: Do we need the include here? It works also without and maybe is better not to restrict that further...?

run-coverage
QuLogic: You can set [`[paths]` in coverage config](https://coverage.readthedocs.io/en/coverage-4.2/config.html#config-paths), then run `coverage combine` to flatten the paths into the first canonical path.

zuphilip: Thank you! Yes, that seems to work. I guess that the PATH construct can be deleted then as well...

PR #300, opened 2018-03-16
complement of readme file

xS-Sx:


README.md
```diff
@@ -19,7 +19,14 @@ trace by default since it seems to confuse too many users).
...
 Installing
 ----------
-
+Preparation: Install git(if not installed); Create working directory to save the source; Pull the respository
+ 
+    $ apt-get install git
+    $ mkdir workingcopy/
+    $ cd workingcopy/ 
+    $ git init .
+    $ git pull https://github.com/tmbdev/ocropy
```
zuphilip: That seems too complicated and `git clone` should be enough for everything here.

kba:
Since this is GitHub, `git clone` is implied for all the install-from-source-steps IMHO.

But if #319 lands, we should consider adding steps to install with pip. In @OCR-D we've been publishing a fork to pypi, so you can do

```
pip install ocrd-fork-ocropy
```

without cloning and in fact use ocropy [as a dependency in other projects](https://github.com/OCR-D/ocrd_ocropy/blob/master/requirements.txt).

zuphilip:
Yes, the `pip` looks really simple. But will this also download the models? Is it clear where I have to save the new models if I have installed `ocropy` with pip locally or globally? (Maybe, we have to add more documentation then or can even provide some sort of helper script for such cases...)

kba:
Users do have to save models to one of the places ocropy searches for them. They can do that manually (wget), use https://github.com/kba/ocr-models-client or we could bundle a script like kraken does.

PR #312, opened 2018-08-30
Output UTF8-HTML-files by default and show args.minscale when skipping args 

NormanTUD:


zuphilip:
@NormanTUD Looks okay in general. How about the overview page in `ocropus-visualize-results`, add the UTF8 part there as well?

NormanTUD:
Also use UTF8-Output for ocropus-visualize-results

zuphilip:
Ah, sorry for the delay. I think here before this block you have to include a line about UTF8 as well: https://github.com/tmbdev/ocropy/blob/65b14307f1dbd3b55006457f6babebdf20cdde9d/ocropus-visualize-results#L102-L107

PR #315, opened 2018-10-29
Fix arguments for function sigmoid

stweil:
Error reported by LGTM:

    Call to function sigmoid with too many arguments;
    should be no more than 1.

Fix it by using the same pattern as in previous calls of that function.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/OLD/mlinear.py
+++ b/OLD/mlinear.py
@@ -298,7 +298,7 @@ class LinKernelClassifier:
         if self.linear:
             pred = dot(make2d(data),self.M.T)
         else:
-            pred = sigmoid(make2d(data),self.M.T)
+            pred = sigmoid(dot(make2d(data),self.M.T))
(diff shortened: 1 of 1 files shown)


stweil:
This is an untested change, so I am not sure that this is the right fix. Nor do I know whether that part of the code is used at all.

See https://lgtm.com/projects/g/tmbdev/ocropy/alerts for details.

stweil:
Then maybe they can be removed?

amitdo:
I suggest to ignore/blacklist 'OLD' in LGTM.

zuphilip:
> Then maybe they can be removed?

Maybe, some of it, other parts could be still used for inspiration to rewrite currently missing functionalities.

> I suggest to ignore/blacklist 'OLD' in LGTM.

This sounds reasonable for me.

PR #322, opened 2019-01-13
Bump version to 1.3.4

zuphilip:


zuphilip:
Version 1.4.0 would also be fine and possibly better reflect the amount of changes.

But I am not sure if we should go here for a version 2, because there is also ocropus2 and ocropus3 which could lead to confusions. Let us maybe freeze the 1 and just count on the second place higher and higher. Thus, your python3-compatible commit should lead to version 1.5.0. What do you think?

kba:
I had the same thought (ocropus2 and ocropy 2.0.0) but reasoned the changes in API (imageio, numpy etc) merited the major bump.  But it really does not matter that much I think, minor bump is fine too.

PR #339, opened 2020-12-03
Update README.md

softana:
Change "source activate" to "conda activate ocropus_env" because the former is cmd.exe syntax and may confuse a linux-based or WSL system.


kba (approved):
Thanks!

